PR #24, opened 2024-07-25
Optimize images in intro and get started

manototh:
Optimize using https://compresspng.com/


dominicchapman (approved):

dominicchapman:
I guess there would be a command-line tool that could do this optimization as part of build step, maybe? Worth checking some other repos for inspo.

manototh:
> I guess there would be a command-line tool that could do this optimization as part of build step, maybe? Worth checking some other repos for inspo.

@dominicchapman Great idea! I'll see what's out there.

PR #28, opened 2024-07-30
fix: Remove tabular format (for now)

bahlo:
This is not (yet) for public use


manototh:
@bahlo Thanks. Is this a breaking change to an existing endpoint? Should we make an announcement?

manototh (approved):

manototh:
@bahlo I understand this is something we shouldn't have documented. Could you please merge? I'm happy with the change.

PR #30, opened 2024-08-05
Add AWS Lambda Using OTel

manototh:


mint.json
```diff
@@ -102,6 +102,7 @@
           "pages": [
             "send-data/cloudfront",
             "send-data/cloudwatch",
+            "send-data/aws-dot",
```
islameldigwi: I believe this should be `adot` not `dot`

mint.json
```diff
@@ -102,6 +102,7 @@
...
             "send-data/aws-firelens",
             "send-data/aws-iot-rules",
             "send-data/aws-s3",
```
islameldigwi: not related to this PR but noticed inconsistency. This is called `aws-s3` but `cloudwatch` is without `aws-` prefix.

nit pick: its usually called [Amazon S3](https://aws.amazon.com/pm/serv-s3/?gclid=Cj0KCQjw8MG1BhCoARIsAHxSiQkn-ZvYD22N6MboXpmDbko_rHPvZebjoLjNze8IsMluCO45XUoJAnEaAmcSEALw_wcB&trk=3cda67b9-5fb7-4d3f-84e8-40b544661f21&sc_channel=ps&ef_id=Cj0KCQjw8MG1BhCoARIsAHxSiQkn-ZvYD22N6MboXpmDbko_rHPvZebjoLjNze8IsMluCO45XUoJAnEaAmcSEALw_wcB:G:s&s_kwcid=AL!4422!3!645208988806!e!!g!!amazon%20s3!19580264380!143903638703). Not sure though why AWS uses Amazon instead of AWS sometimes.

mint.json
manototh: Thanks for this. The official name is `AWS Distro for OpenTelemetry` so we either abbreviate it as `AWS DOT` or `ADOT`. But `AWS ADOT` isn't correct.

mint.json
manototh: Well spotted! It would be good to have consistent naming in the file names, but I don't think it's worth changing the URL and add redirects. I'll pay attention to this in the text, though.

mint.json
islameldigwi: makes sense, but from the first look just `dot` feels weird, I haven't seen it written anywhere like that (in their docs). In this [page](https://aws-otel.github.io/docs/introduction) they always write `ADOT`. Maybe the url can just do `/send-data/ADOT` without aws part?

thesollyz (approved):

manototh:
@SollyzDev Could you please check this again? I've substantially rewritten the page.
https://axiom-mano-aws-dot.mintlify.app/send-data/aws-lambda-dot

Also, on the page about Axiom Lambda Extension, is it necessary to add the env vars for token and dataset if you [install it via the AWS CLI](https://axiom-mano-aws-dot.mintlify.app/send-data/aws-lambda#install-with-aws-cli)?

thesollyz:
> @SollyzDev Could you please check this again? I've substantially rewritten the page. https://axiom-mano-aws-dot.mintlify.app/send-data/aws-lambda-dot
> 
> Also, on the page about Axiom Lambda Extension, is it necessary to add the env vars for token and dataset if you [install it via the AWS CLI](https://axiom-mano-aws-dot.mintlify.app/send-data/aws-lambda#install-with-aws-cli)?

Yes, they have to be put to defined in the main function environment variables, not sure if its possible to define them in the same command as adding the layer.

Also, I noticed `Create an account on AWS Cloud` step, and it didn't reason with me. Perhaps users who visit this page already have lambdas and already have an AWS account. In my opinion, it would be better if we replace this step that guides users to define credential environment variables.

manototh:
@SollyzDev Thanks a lot for this. I've added the step about the environment variables.

As for the prerequisites, you're right that some users have satisfied the AWS account prerequisite. But this is true of all prerequisites. I think it's reasonable to assume that users understand that from the list of prereqs they only need to work with the ones they haven't already satisfied.

PR #39, opened 2024-08-19
Revise Flows

manototh:


manototh:
@dominicchapman Thanks a lot for tending to the Flows docs! ❤️ 

I've revised the wording in light of the new terminology (route instead of transfer) and added a bit more context to the standalone destination pages. Thanks in advance for reviewing this when you have the time.

dominicchapman (approved):

PR #50, opened 2024-08-28
extract takes an integer not a string

gordallott:
noticed this error in our docs, just a small fix


manototh (approved):

manototh:
@gordallott Great, thank you so much!

PR #51, opened 2024-08-28
Add anomaly monitors

manototh:
Sources:
https://www.notion.so/axiomhq/Time-series-Anomaly-Detection-5a023d46fd47488f92f6384bddbf5dd1
https://www.notion.so/axiomhq/Anomaly-detection-c14b34a47847404982b76e537dd1c394
https://app-previews-r00fq9p31.axiom.dev/axiomers-ft83/monitors/cdLun5smddW3pONew3


monitor-data/monitors.mdx
```diff
@@ -58,6 +59,33 @@ To create a threshold monitor, follow these steps:
 
 You have created a threshold monitor, and Axiom alerts you when the results from your query cross the threshold.
 
+## Anomaly monitors
+
+Anomaly monitors allow you to aggregate your event data and compare the results of this aggregation to what can be considered normal for the query. When the results are too much above or below the value that Axiom expects based on the event history, the monitor enters the alert state. The monitor remains in the alert state until the results no longer deviate from the expected value. An anomaly monitor sends you a notification each time it enters or exits the alert state.
```
benenharrington:
```suggestion
Anomaly monitors allow you to aggregate your event data and compare the results of this aggregation to what can be considered normal for the query. When the results are too much above or below the value that Axiom expects based on the event history, the monitor enters the alert state. The monitor remains in the alert state until the results no longer deviate from the expected value. This can happen without the results returning to their previous level if they stabilize around a new value. An anomaly monitor sends you a notification each time it enters or exits the alert state.
```

monitor-data/monitors.mdx
```diff
@@ -58,6 +59,33 @@ To create a threshold monitor, follow these steps:
...
+To create an anomaly monitor, follow these steps:
+
+1. Click the **Monitors** tab on the top of the page, and then click **New monitor**.
+2. Click **Anomaly monitor**.
+3. Name your monitor and add a description.
+4. Configure the monitor using the following options:
+    - The comparison operator is the rule to apply when comparing the results to the expected value. The possible values are **above**, **below**, and **above or below**.
+    - The tolerance factor controls the sensitivity of the monitor. It defines how much deviation from the expected value to accept without triggering the monitor. It’s a range above and below the expected value. The higher the tolerance factor, the wider this range. When the results of the aggregation stay within this range, the monitor doesn’t trigger. When the results of the aggregation cross this range, the monitor triggers. The tolerance factor can be any positive numeric value.
+    - The frequency is how often the monitor runs. This is a positive integer number of minutes.
+    - The range is the time range for your query. This is a positive integer number of minutes. The end time is the time the monitor runs.
```
benenharrington: I don't understand the mention of end time here. Adding an additional line about the range. Two separate suggestions!

```suggestion
    - The range is the time range for your query. This is a positive integer number of minutes. A longer time range allows the anomaly monitor to consider a larger number of datapoints when calculating the expected value.
```

monitor-data/monitors.mdx
```diff
@@ -58,6 +59,33 @@ To create a threshold monitor, follow these steps:
...
+    - You can group by attributes when defining your query. By default, your monitor enters the alert state if any of the values returned for the group-by attributes deviate from the expected value, and remains in the alert state until none of the values returned deviates from the expected value. To trigger the monitor separately for each group that deviates from the expected value, enable **Notify by group**. At most one trigger notification is sent per monitor run. This option only has an effect if the monitor’s query groups by a non-time field.
+    - Toggle **Require seasonality** to compare the results to seasonal patterns in your data. For example, your event history displays a normal pattern of a high number of events during the day and a low number of events during the night. Without accounting for seasonality, the monitor compares deviations to the average and triggers even if the deviations are due to the normal seasonal variation. By toggling **Require seasonality**, the monitor compares the results to the same time of the previous day or week and only triggers if the results deviate from the expected seasonal pattern.
```
benenharrington:
```suggestion
    - Toggle **Require seasonality** to compare the results to seasonal patterns in your data. For example, your query produces a time series that increases at the same time each morning. Without accounting for seasonality, the monitor compares to recent results only. By toggling **Require seasonality**, the monitor compares the results to the same time of the previous day or week and only triggers if the results deviate from the expected seasonal pattern.
```

benenharrington (commented):
This is fantastic @tothmano , thank you very much for getting this together so quickly. I'd like to take the time to add an example monitor that is of anomaly type and then I think we're good to go

benenharrington:
I added an example to the branch `benen/anomaly-monitors-example`. Please merge it in if you're happy with it!

manototh:
> This is fantastic @tothmano , thank you very much for getting this together so quickly. I'd like to take the time to add an example monitor that is of anomaly type and then I think we're good to go

@benenharrington Thanks a lot for your additions! All looks good from my side.

benenharrington (approved):

PR #55, opened 2024-08-29
Add audit log

manototh:


reference/audit-logs.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+- Track who has accessed the Axiom platform.
+- Track organization access over time.
+- Track data access over time.
+
+Audit logs also make it easier to manage your Axiom organization. They allow you to do the following, among others:
+
+- Track changes made by your team to your observability posture.
+- Track monitoring performance.
+- Track billable usage.
```
benenharrington: My bad for outdated Notion page, but this one has not made it in and will likely eventually be surfaced in a different dataset

reference/audit-logs.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+Audit logs are available to all users. Enterprise customers can query audit logs for the full time range. Other customers can query audit logs for the previous three days.
+
+## Explore audit logs
+
+1. Go to the **Explore** tab, and then click **Builder**.
+2. In the list of datasets, click **axiom-audit**.
```
benenharrington: Audit log will not appear in a user's list of datasets. They'll need to write an APL query referencing the reserved 'axiom-audit' dataset to access

reference/audit-logs.mdx
```diff
@@ -0,0 +1,69 @@
...
+- **removeUserFromOrg** means that a user removed another user from an organization.
+- **runAplQuery** means that a user ran an APL query.
+- **sendMessage** means that Axiom sent a message to a channel defined by a notifier.
+- **sendUserDeletedEmail** means that a user was deleted and Axiom sent out a notification email about the deletion.
+- **streamDataset** means that a user viewed the stream of a dataset.
+- **trimDataset** means that a user trimmed a dataset.
+- **updateDashboard** means that a user updated a dashboard.
+- **updateMonitor** means that a user deleted a monitor.
+- **updateUserSettings** means that a user updated their user settings.
+- **updateVirtualField** means that a user updated a virtual field.
```
benenharrington: I worry that we'll struggle to keep up with these as more are added (indeed, this current list is not complete). I think the value of action should mostly be self-documenting?

manototh: @benenharrington Thanks a lot for your review. I agree this is a maintenance hassle. I don't believe in self-documenting field names, but we can make the worst offenders easier to understand.

What do you think about renaming the following?
sendMessage --> sendMessageToNotificationChannel
sendUserDeletedEmail --> deleteUserSendNotificationEmail

Then we won't need this list, hopefully.

manototh: @benenharrington Maybe we can include more context in the `message` field? Then we won't need this list. I'm happy to come up with a PR in the axiom repo if you think it's worthwhile.
Is there anything else I should sort out on this page?

benenharrington: Happy with any changes to action names required to avoid documentation 👍 

No need to expand the message field, I think it would likely be quite a lot of work to have something meaningful here for all actions. Can iterate if we start to get questions

benenharrington: > Is there anything else I should sort out on this page?

Nope!

benenharrington (approved):

PR #62, opened 2024-09-06
Fix syntax in query examples

manototh:


restapi/query.mdx
```diff
@@ -303,7 +303,7 @@ curl -X 'POST' 'https://api.axiom.co/v1/datasets/_apl?format=tabular' \
 -H 'Accept-Encoding: gzip' \
 -H 'Content-Type: application/json' \
 -d '{
-      "apl": "github-issue-comment-event | where repo startswith \"kubernetes/\" | where actor !endswith \"[bot]\" | summarize dcount(actor) by bin_auto(_time)",
+      "apl": "\"github-issue-comment-event\" | where repo startswith \"kubernetes/\" | where actor !endswith \"[bot]\" | summarize dcount(actor) by bin_auto(_time)",
```
mhr3: missing brackets

restapi/query.mdx
```diff
@@ -318,7 +318,7 @@ curl -X 'POST' 'https://api.axiom.co/v1/datasets/_apl?format=tabular' \
 -H 'Accept-Encoding: gzip' \
 -H 'Content-Type: application/json' \
 -d '{
-      "apl": "github-issues-event | where (actor !endswith \"[bot]\" and repo startswith \"cockroachdb/\" and actor !~ \"cockroach-teamcity\") | summarize topk(actor, 5) by bin_auto(_time), action",
+      "apl": "\"github-issues-event\" | where (actor !endswith \"[bot]\" and repo startswith \"cockroachdb/\" and actor !~ \"cockroach-teamcity\") | summarize topk(actor, 5) by bin_auto(_time), action",
```
mhr3: ditto

mhr3 (commented):
Also noticed the Authorization header is using `-H 'Authorization: Bearer $API_TOKEN'` which is wrong, if the token was stored in an env, the use of `'` would prevent expanding it and a literal `$API_TOKEN` would be sent as a token.

restapi/query.mdx
ted-gould: The parens in the where aren't really needed.

restapi/query.mdx
manototh:
```suggestion
      "apl": "[\"github-issue-comment-event\"] | where repo startswith \"kubernetes/\" | where actor !endswith \"[bot]\" | summarize dcount(actor) by bin_auto(_time)",
```

restapi/query.mdx
manototh:
```suggestion
      "apl": "[\"github-issues-event\"] | where actor !endswith \"[bot]\" and repo startswith \"cockroachdb/\" and actor !~ \"cockroach-teamcity\" | summarize topk(actor, 5) by bin_auto(_time), action",
```

manototh:
@mhr3 @ted-gould Thanks a lot both! I think this should be fine now. Could you please take another look?

manototh:
> Also noticed the Authorization header is using `-H 'Authorization: Bearer $API_TOKEN'` which is wrong, if the token was stored in an env, the use of `'` would prevent expanding it and a literal `$API_TOKEN` would be sent as a token.

Well spotted. Corrected. I would like to move towards using `API_TOKEN` as placeholder everywhere in the docs. This is in line with the Google Developer style guide and it leaves the issue of using env vars to the user.

mhr3 (approved):

PR #64, opened 2024-09-10
Update dashboards and filter after UI update

manototh:


query-data/dashboards.mdx
```diff
@@ -29,13 +29,15 @@ Select a private or shared dashboard from this tab. A private dashboard is one t
...
+1. Click **+ Add a chart** in the top right corner.
 
     <Frame caption="New chart button">
         <img src="/doc-assets/shots/your-first-query-1.png" alt="New chart button" />
     </Frame>
 
-3. Choose one of the following:
+1. Choose the chart type from the list. For more information, see [Chart types](#chart-types).
+
+1. Choose one of the following:
```
c-ehrlich, line 28: seems like all of these are numbered `1.`?

query-data/dashboards.mdx
```diff
@@ -29,13 +29,15 @@ Select a private or shared dashboard from this tab. A private dashboard is one t
 
     You see an empty dashboard.
 
-2. Click **+ Add a chart** in the top right corner.
+1. Click **+ Add a chart** in the top right corner.
 
     <Frame caption="New chart button">
         <img src="/doc-assets/shots/your-first-query-1.png" alt="New chart button" />
```
c-ehrlich: IMO it's better to train users to use the "+" button in the top right corner from the start. The "Add a chart" card is only visible in new dashboards without charts, and we will probably revise the empty dashboard exprience very soon.

c-ehrlich (approved):

query-data/dashboards.mdx
c-ehrlich: oh, i see that it figures out the numbers on its own :)

query-data/dashboards.mdx
manototh: True that. Updated along with some other things.

PR #68, opened 2024-09-13
fix: Add required fields to tabular result

bahlo:
* Mark AplResult.request as not-required as it's not present on tabular results.


Licenser (approved):
🥳

PR #83, opened 2024-09-20
Update v2 swagger file from make server-docs command

manototh:


Rambatino (approved):
If you're happy all is good :)

PR #84, opened 2024-09-23
Add delete data endpoint

manototh:
Preview: https://axiom-mano-add-delete-endpoint.mintlify.app/restapi/endpoints/deleteData


manototh:
Not to be publicly documented

PR #94, opened 2024-10-03
Fix styling of inline icons

manototh:
Inline icons are rendered incorrectly after some update at mintlify
image


manototh:
No longer necessary. Bug fixed on Mintlify's side.

PR #96, opened 2024-10-04
Add overviews, logoIds and missing ingest options

thesollyz:
- [X] Added "overview" in mdx files to show short and concise descriptions
- [X] Added logoIds
- [X] Added Zap, Apex & Logrus


manototh:
@dasfmi Thanks a lot for this! I love how this looks in the app! Really good job. I also like how aligned the grouping labels in the app and the docs nav are.

As for creating a dependency between the docs and the app in this PR, I think I would like to learn more about the benefits of this automated approach, please. What is the advantage of generating the `ingest-options.json` file based on the docs as opposed to keeping this list in the app repo?

From what I see, there might be some cons:
- The content of the `ingest-options.json` file is incorporated in the source HTML of every page on Mintlify prod. It's not a huge file but I'm wary of adding unnecessary complexity to docs pages.
- The `scripts/generate-ingest-options.mjs` currently parses the `mint.json` file for every page within the Send Data section. This creates a dependency between the app code and the docs file structure and nav structure which I would not be comfortable with.
- What happens if we add a new page to the Send Data section but there is no corresponding logo in the app repo? We need to update the app repo so we need to manually edit both repos anyway.
- In this PR, the logic for rending popular items in the app is stored in the docs repo where this information doesn't do anything. To me, it seems like an unnecessary dependency.

There might be an angle I don't see here and I would be grateful if you can show me that. At the first glance, this automation feels a bit overengineered which might unnecessarily tie our hands in the docs repo in the long run, especially because the point of moving to Mintlify was to remove some of the dependency and maintenance burden for which we would need help from outside the Product team.

If the rationale behind this automation is that the links in the docs might change, I can assure you that all links will be maintained. The file structure might change but I always add redirects to avoid broken links.

thesollyz:
hi @mano, thanks for the feedback. Will try to answer your questions.

> What is the advantage of generating the ingest-options.json file based on the docs as opposed to keeping this list in the app repo?

**The problem**: We have a hard coded list in the app repo and it was outdated and out of sync with the doc. No one was taking care of that list nor updating it. So when new items are added, app can have them too.


**Solution**:
In order to overcome this, we thought of automating the update process, whenever something in the docs changes, it would change in the app as well.

> The content of the ingest-options.json file is incorporated in the source HTML of every page on Mintlify prod. It's not a huge file but I'm wary of adding unnecessary complexity to docs pages.

The file is just in the repo, it shouldn't be served anywhere, are you sue its in every HTML?

> The scripts/generate-ingest-options.mjs currently parses the mint.json file for every page within the Send Data section. This creates a dependency between the app code and the docs file structure and nav structure which I would not be comfortable with.

The app only cares about the output (a JSON file). The script lives in the same repo as nav. The script uses same values as mint.json.

> What happens if we add a new page to the Send Data section but there is no corresponding logo in the app repo? We need to update the app repo so we need to manually edit both repos anyway.

The plan I have is that when docs change, a PR would be created automatically in the app repo, then tested/checked and merged, if an icon is missing, then it can be added.

As we want app and docs to be in-sync, I believe some kind of collaboration will have to be made.

> In this PR, the logic for rending popular items in the app is stored in the docs repo where this information doesn't do anything. To me, it seems like an unnecessary dependency.

I don't mind moving this to app, but we wanted a single source of truth for this data. Wouldn't it be beneficial to have a Popular section in our docs, or maybe sort send-data by popularity ? anyhow, I am flexible with removing this.

> If the rationale behind this automation is that the links in the docs might change, I can assure you that all links will be maintained. The file structure might change but I always add redirects to avoid broken links.

The rationale behind this is to keep app in-sync with docs, when new items are added, they should appear in app as well.

thesollyz:
update: we are going to remove the script from this repo, and it will live inside the app, but the docs will still be the source of truth.

manototh:
@dasfmi Thanks a lot for the explanation. I try to reply here. Also, please accept my apology for not raising these issues earlier when you were still just planning this. I didn't understand the full scope of the project.

> hi @mano, thanks for the feedback. Will try to answer your questions.
> 
> > What is the advantage of generating the ingest-options.json file based on the docs as opposed to keeping this list in the app repo?
> 
> **The problem**: We have a hard coded list in the app repo and it was outdated and out of sync with the doc. No one was taking care of that list nor updating it. So when new items are added, app can have them too.
> 
> **Solution**: In order to overcome this, we thought of automating the update process, whenever something in the docs changes, it would change in the app as well.

I see your point. It's a worthy goal. What I worry about is whether we create more issues with these dependencies. I mean, we don't add new data sources that often. I would be more comfortable to commit to a process where I manually update things in both repo. I'm more worried about changing something in the docs repo and then inadvertently breaking something in the app.

> > The content of the ingest-options.json file is incorporated in the source HTML of every page on Mintlify prod. It's not a huge file but I'm wary of adding unnecessary complexity to docs pages.
> 
> The file is just in the repo, it shouldn't be served anywhere, are you sue its in every HTML?

Yes, before your last commit it was hard-coded in the HTML source of every page. No biggie, but unnecessary.

> > The scripts/generate-ingest-options.mjs currently parses the mint.json file for every page within the Send Data section. This creates a dependency between the app code and the docs file structure and nav structure which I would not be comfortable with.
> 
> The app only cares about the output (a JSON file). The script lives in the same repo as nav. The script uses same values as mint.json.

Yes, but the script takes whole Send Data section from the docs without filtering. What happens if I want to add an Overview page to Collectors, for example? I don't want to worry about breaking the app each time I change this section.

> > What happens if we add a new page to the Send Data section but there is no corresponding logo in the app repo? We need to update the app repo so we need to manually edit both repos anyway.
> 
> The plan I have is that when docs change, a PR would be created automatically in the app repo, then tested/checked and merged, if an icon is missing, then it can be added.
>
> As we want app and docs to be in-sync, I believe some kind of collaboration will have to be made.

Happy to collaborate, sure. But I would be happier with a less fancy but safer manual process. What is the point of the automation if we need to add things like icons manually anyway?
 
> > In this PR, the logic for rending popular items in the app is stored in the docs repo where this information doesn't do anything. To me, it seems like an unnecessary dependency.
> 
> I don't mind moving this to app, but we wanted a single source of truth for this data. Wouldn't it be beneficial to have a Popular section in our docs, or maybe sort send-data by popularity ? anyhow, I am flexible with removing this.
> 
> > If the rationale behind this automation is that the links in the docs might change, I can assure you that all links will be maintained. The file structure might change but I always add redirects to avoid broken links.
> 
> The rationale behind this is to keep app in-sync with docs, when new items are added, they should appear in app as well.

Sure, the two lists should be in sync. I am happy to commit to updating both repos whenever I add a new entry in the Send Data section in the docs. What I'm not comfortable with is using the docs file structure as the source of truth. Indiscriminately taking all entries from the Send Data section doesn't work for me. I don't want to break the app by adding pages and restructuring the left nav, and I don't want this dependency to tie my hands to do that. So moving the script to the app repo doesn't solve the issue, unfortunately.

bahlo:
@tothmano There were a lot of arguments _for_ having docs as a single source of truth when we talked about the SDK doc location. I hear that you're worried about breaking the app and creating strong dependencies; the way we implemented it now is automatically creating a pull request that needs to be acknowledged by an engineer, so there's no way you can break the app.

manototh:
> @tothmano There were a lot of arguments _for_ having docs as a single source of truth when we talked about the SDK doc location. I hear that you're worried about breaking the app and creating strong dependencies; the way we implemented it now is automatically creating a pull request that needs to be acknowledged by an engineer, so there's no way you can break the app.

@bahlo Thanks a lot for hearing my concerns. That sounds promising indeed. What do I need to watch out for then if we merge this? Not to touch the frontmatter tags added in this PR?

Also, should we keep the popularity and logo information in the docs? I don't mind these tags much, but they have an effect in the app, not in the docs.

bahlo:
@tothmano You can do whatever, it's your repo. Frontend will ping you if something's off with the ingest options and then we can resolve it together. I get that some of these keys can seem out-of-place as they're only used in the app, but I'm still positive this is the best way forward.

README.md
```diff
@@ -54,6 +54,14 @@ Run the following in your terminal to check the style of all files in the curren
...
 ```
 
+## Generate ingest-options json file
+
+Run the following in your terminal to generate the ingest-options.json file:
+
+```
+npm run generate-ingest-options
+```
+
```
manototh:
```suggestion
```

manototh:
> @tothmano You can do whatever, it's your repo. Frontend will ping you if something's off with the ingest options and then we can resolve it together. I get that some of these keys can seem out-of-place as they're only used in the app, but I'm still positive this is the best way forward.

@bahlo Thank you for accommodating my concerns. I appreciate this a lot. I feel much safer knowing there is a human element in this process. It seems to me the best compromise.

I didn't mean to send the message that this is "my" repo or that I don't want to make any changes to it that will make everyone's life easier. I own the repo and it's my responsibility, sure, but it's an open repo, internally and externally. Naturally, I am willing to accept limitations and dependencies as we integrate the app and the docs better. It's just that in this case I worried that the initial conception meant that a docs change can break the app. I think this is no longer the case.

I would still like to know better what changes to the docs can mess up the list generated by the script. If I know this, I can try to avoid this or at least work with you to update this process before the fact rather than after breaking it. For example, do additions to the Send Data section trigger the script to generate a new list? If I add an overview page to Collectors, for example, will we need to update the script?

Thanks in advance.

thesollyz:
> @bahlo Thank you for accommodating my concerns. I appreciate this a lot. I feel much safer knowing there is a human element in this process. It seems to me the best compromise.

I am glad we were able to address your concerns and found a solution that works for you and the repo. Absolutely appreciated work on your side 👍 

> I would still like to know better what changes to the docs can mess up the list generated by the script. If I know this, I can try to avoid this or at least work with you to update this process before the fact rather than after breaking it. For example, do additions to the Send Data section trigger the script to generate a new list? If I add an overview page to Collectors, for example, will we need to update the script?

Right now, when there is changes, an automatic PR will be created with the changes. If doesn't break anything, then good. If it fails or something is buggy or missing, the script can be updated to fix it. So its still a manual process (kind of) but parts of it are automated to make it easier to keep app-docs in sync.

For things like an overview page, we can either:
- Exclude it in the script
- Introduce a flag variable that tells the script to ignore the file that includes it

manototh (approved):

manototh:
@bahlo @dasfmi Thanks both! I appreciate your flexibility and your patience with this. Looking good to me!

PR #98, opened 2024-10-07
Add pie charts

manototh:


doc-assets/shots/pie-chart-apl.png
ImLunaHey, line 1: let's drop the `bin_auto` part, `summarize count() by status` would be a better example.

ImLunaHey (approved):

bahlo (approved):

PR #103, opened 2024-10-07
aws: Cloudwatch Forwarder terraform module

thesollyz:
We need to explain the different components (Forwarder, Subscriber, Unsubscriber, Listener) without repeating ourselves in Cloudformation and Terraform.


apps/vercel.mdx
```diff
@@ -260,3 +263,7 @@ module.exports = withAxiom({
 ```
 
 For more configuration options, see the [documentation in the next-axiom GitHub repository](https://github.com/axiomhq/next-axiom).
+
+## Send logs from Vercel preview deployments
```
manototh: I've moved this section because it seems to be related to the Vercel integration rather than Next.js.

send-data/cloudwatch.mdx
```diff
@@ -28,12 +29,88 @@ Axiom CloudWatch Forwarder includes templates for the following CloudFormation s
...
+
+</Step>
+<Step>
+In your terminal, go to the folder of your main Terraform file, and then run `terraform init`.
+</Step>
+<Step>
+Run `terraform plan` to check the changes, and then run `terraform apply`.
+</Step>
+<Step>
+To remove subscriptions from the Forwarder, use the Unsubscriber module.
```
manototh: How to use this? Just create a new module in the Terraform file? Why can't users create this earlier?

send-data/cloudwatch.mdx
```diff
@@ -28,12 +29,88 @@ Axiom CloudWatch Forwarder includes templates for the following CloudFormation s
...
+module "forwarder" {
+  source           = "axiomhq/axiom-cloudwatch-forwarder/aws//modules/forwarder"
+  axiom_dataset    = "DATASET_NAME"
+  axiom_token      = "API_TOKEN"
+  prefix           = "axiom-cloudwatch-forwarder"
+}
+```
+
+<Replacement />
+Alternatively, use Axiom’s provider to create a dataset.
```
manototh: How can users create a dataset with Axiom's provider?

apps/vercel.mdx
islameldigwi: looks like my commits got mixed in this PR by mistake.

send-data/cloudwatch.mdx
islameldigwi: So, this shouldn't be part of initial setup as this removes the subscriptions from the Forwarder, meaning, it disabled the log forwarding for the selected log groups. Users might not need to do that at all, and if they do it won't be during setup.

send-data/cloudwatch.mdx
islameldigwi: This is the docs for creating datasets using Terraform => https://axiom.co/docs/apps/terraform#create-dataset

apps/vercel.mdx
islameldigwi: removed them from this PR.

send-data/cloudwatch.mdx
manototh: Ah, you mean the Axiom Terraform provider!

manototh (approved):

PR #105, opened 2024-10-09
Add best practices section

thecraftman:
- best practices for ingesting data

- best practices on using AWS Lambda


send-data/aws-lambda.mdx
```diff
@@ -199,3 +199,15 @@ To turn off Amazon CloudWatch logging, deny the Lambda function access to Amazon
...
+
+## Best practices using AWS Lambda
+
+When using AWS Lambda for ingestion, consider these additional recommendations:
+
+- Axiom Lambda extension: Use the [Axiom Lambda extension](/docs/send-data/aws-lambda) for efficient log collection directly from your Lambda functions.
+
+- Minimize direct API calls: Avoid making individual API calls for each Lambda invocation. Instead, use the [Lambda extension](/docs/send-data/aws-lambda) or aggregate logs before sending.
+
+- Terminate early: In Lambda functions, prioritize quick execution and termination. Avoid waiting for ingestion confirmation within the function itself.
```
pecigonzalo: While this is correct, I think it might cause confusion as to send and exit without waiting, which might lead to lost logs.

send-data/aws-lambda.mdx
```diff
@@ -199,3 +199,15 @@ To turn off Amazon CloudWatch logging, deny the Lambda function access to Amazon
...
 
 For testing purposes, set the `PANIC_ON_API_ERR` environment variable to `true`. This means that the Axiom Lambda Extension crashes if it can’t connect to Axiom.
+
+## Best practices using AWS Lambda
+
+When using AWS Lambda for ingestion, consider these additional recommendations:
+
+- Axiom Lambda extension: Use the [Axiom Lambda extension](/docs/send-data/aws-lambda) for efficient log collection directly from your Lambda functions.
+
+- Minimize direct API calls: Avoid making individual API calls for each Lambda invocation. Instead, use the [Lambda extension](/docs/send-data/aws-lambda) or aggregate logs before sending.
```
pecigonzalo: These feel like the same thing really, both explanations are good and valid, but ultimately are the same right?

send-data/aws-lambda.mdx
```diff
@@ -199,3 +199,15 @@ To turn off Amazon CloudWatch logging, deny the Lambda function access to Amazon
...
+
+- Terminate early: In Lambda functions, prioritize quick execution and termination. Avoid waiting for ingestion confirmation within the function itself.
+
+- Alternative routing option: If you aren‘t using the Axiom Lambda layer, consider routing to an internal gateway that handles batching and processing. For example: Lambda -> CloudWatch Logs (short retention) -> Log Forwarder. This approach allows for efficient log aggregation and processing before sending to Axiom.
```
pecigonzalo:
```suggestion
- Alternative routing option: If you aren‘t using the Axiom Lambda layer, consider routing to an internal gateway that handles batching and processing. For example: Lambda -> CloudWatch Logs (short retention) -> [Log Forwarder](https://github.com/axiomhq/axiom-cloudwatch-forwarder). This approach allows for efficient log aggregation and processing before sending to Axiom.
```

send-data/ingest.mdx
```diff
@@ -284,3 +284,13 @@ All events stored in Axiom must have a `_time` timestamp field. If the data you
 </Note>
 
 If you include the `_time` field in the ingested data, ensure the `_time` field contains timestamps in a valid time format. Axiom accepts many date strings and timestamps without knowing the format in advance, including Unix Epoch, RFC3339, or ISO 8601.
+
+## Best practices for ingesting data
+
+When ingesting data into Axiom, follow these best practices to optimize performance and reliability:
+
+- Batch events: Instead of sending individual events, group multiple events into a single request. This reduces the number of API calls and improves overall throughput.
```
pecigonzalo: Does this refer to doing it on the app itself or to configuring your log forwarder to do it? If the later :+1:, if the former, its complicated as there are several considerations there :)

dominicchapman: @pecigonzalo I’m not following "on the app itself". Can you help?

send-data/ingest.mdx
```diff
@@ -284,3 +284,13 @@ All events stored in Axiom must have a `_time` timestamp field. If the data you
...
+
+- Use Compression:  Enable gzip, zstd compression for your requests to reduce bandwidth usage and potentially improve response time.
+
+- Handle rate limiting and errors: Implement exponential backoff and retry logic to handle 429 (Too Many Requests) and 503 (Service Unavailable) errors. If ingestion consistently fails after retries, implement a fallback mechanism to store events locally or in cold storage for later processing.
```
dominicchapman: This could be intimidating for hobbyist or less-technical users. We should be clear that our libraries take care of this... assuming they do (and they should, if they don’t :-) )

send-data/ingest.mdx
pecigonzalo: Sure! 

When we say batch the logs before sending here, are we saying to use some system or pattern to batch them? eg. `app <- otel-collector -> axiom` in which case otel-collector does the batching

or we are saying, in your application buffer the logs and then write them to axiom as a batch?

thecraftman:
moved here: https://github.com/axiomhq/docs/pull/125

PR #109, opened 2024-10-11
Remove go version from CLI page

bahlo:
The supported version will go out of sync again if we updated it.


manototh:
@bahlo Thanks! Should we say the latest stable version then?

manototh (approved):

PR #137, opened 2024-10-31
Update query costs

manototh:


reference/query-hours.mdx
```diff
@@ -19,6 +18,18 @@ Axiom uses serverless computing to execute queries efficiently. The consumption
 - Time: The duration (in milliseconds) for which the serverless function is running to execute your query.
 - Memory allocation: The amount of memory (in GB) allocated to the serverless function during execution.
 
+## What counts as a query
+
+In calculating query costs, Axiom considers any request that queries your data as a query. For example, the following all count as queries:
+
+- You initiate a query in the Axiom user interface.
+- You query your data with an API token or a personal access token.
+- Your match monitor runs a query to determine if any new events match your criteria.
```
ronoc, line 27: it's not just  a match monitor, it's any type of monitor

ronoc (approved):

reference/query-hours.mdx
manototh: @ronoc Thanks! Yes, that's true. This list isn't meant to be exhaustive. It just gives a list of possible queries. I would leave it as it is.

dominicchapman (approved):
You went a different direction with this to the one I originally had in mind (keeping a _Costs_ page, and introducing other cost-related elements). 

I don’t mind that at all as a way to be clearer about what _this_ content says today, so approving :-)

I would recommend creating an Issue to track follow-up work in the future that captures the Axiom Pricing v0.2 material I shared with you directly in a suitable documentation format.  It could well be that at that point the more exhaustive reference deserves the _Costs_ title once again.

manototh:
@dominicchapman Thanks a lot for this. Agreed, this will be refashioned as a general cost page with the new structure.

PR #148, opened 2024-11-12
Update backreferences in string functions

manototh:
Preview: https://axiom-mano-fix-backreferences.mintlify.app/apl/scalar-functions/string-functions#backreferences

### DIFF
--- a/apl/scalar-functions/string-functions.mdx
+++ b/apl/scalar-functions/string-functions.mdx
@@ -721,10 +721,10 @@ Backreferences match the same text as previously matched by a capturing group. W
 
 ```kusto
 ['github-issues-event']
-| project backreferences = replace_regex(@'observability=(d+)', 'axiom=$1', creator)
+| project backreferences = replace_regex(@'observability=(.+)', 'axiom=$1', creator)
 ```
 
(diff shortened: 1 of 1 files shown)


apl/scalar-functions/string-functions.mdx
```diff
@@ -721,10 +721,10 @@ Backreferences match the same text as previously matched by a capturing group. W
 
 ```kusto
 ['github-issues-event']
-| project backreferences = replace_regex(@'observability=(\d+)', 'axiom=$1', creator)
+| project backreferences = replace_regex(@'observability=(d+)', 'axiom=$1', creator)
```
Rambatino: that's not matching a number though now

manototh: @Rambatino Oh, how dumb of me. I've replaced `d` with `.` so that it makes sense but avoids the `\`. Please let me know if this works.

Rambatino (approved):

apl/scalar-functions/string-functions.mdx
Rambatino: yeah seems good!

PR #150, opened 2024-11-14
Add IP functions section

manototh:


apl/scalar-functions/ip-functions/ipv4-is-private.mdx
```diff
@@ -0,0 +1,109 @@
...
+
+## Use case example
+
+You can use `ipv4_is_private` to filter logs and focus on public traffic for external analysis.
+
+**Query**
+
+```kusto
+['sample-http-logs']
+| extend is_private = ipv4_is_private('192.168.0.1')
```
ted-gould, line 96: I think that a better example here might be using `countif` so:

`| summarize private = countif(ipv4_is_private(client_addr)), total = count() by bin_auto(_time)`

Something like that.

apl/scalar-functions/ip-functions/ipv4-netmask-suffix.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+## Use case example
+
+When analyzing network traffic logs, you can extract the netmask suffix to group or filter traffic by subnets.
+
+**Query**
+
+```kusto
+['sample-http-logs']
+| extend netmask = ipv4_netmask_suffix('192.168.1.1/24')
```
ted-gould, line 78: Hmm, can't come up with a better example here. Perhaps Stefan would have one. (don't know his Github handle)

apl/scalar-functions/ip-functions/parse-ipv4.mdx
```diff
@@ -0,0 +1,143 @@
...
+<Tab title="Log analysis">
+
+You can use the `parse_ipv4` function to analyze web traffic by breaking down user IP addresses into octets.
+
+**Query**
+
+```kusto
+['sample-http-logs']
+| extend ip_octets = parse_ipv4('192.168.1.1')
+| project _time, uri, method, ip_octets
```
ted-gould: So I don't know how to make this into a reasonable example. But for a lot of corps they split up the private IP range into sites. So 10.35.* is Texas and 10.36.* is California, etc. They'd probably filter on 10 being the first digit and then group based on the second. Something like that? Not quite sure on how to make it something easy to understand.

ted-gould (approved):
Looks awesome, just a couple ideas for examples!

manototh:
@ted-gould Many thanks for your review here and for your ideas. I agree with your comments. Ideally, I would like examples that use fields from the Playground. However, we don't have an IP field in any of our Play datasets. So for these functions, we need to resort to these simplistic, hard-coded IP addresses. This is how KQL does it as well. It's not ideal, but unfortunately necessary.

PR #156, opened 2024-11-21
Add short page on event data

manototh:
Preview:
- https://axiom-mano-short-event-data.mintlify.app/getting-started-guide/event-data
- https://axiom-mano-short-event-data.mintlify.app/getting-started-guide/observability


getting-started-guide/event-data.mdx
```diff
@@ -0,0 +1,58 @@
+---
+title: "Events in Axiom"
+description: "This page explains the fundamentals of event data in Axiom."
+sidebarTitle: Events
+---
+
+In Axiom, event data is the atomic unit of activity, encompassing logs, traces, and even product interactions. Each event is simply a structured record—composed of key-value pairs—that captures meaningful interactions or changes in state within a system. While these can appear in various forms, from log lines to complex traces, they usually contain the following:
```
dominicchapman: Let’s call it "timestamped event data" in the first mention, and event data thereafter. I want to ground event data very clearly in time from the outset. Note: This is _not_ described as time _series_ event data. The language choice is intentional, to distance Axiom rightly from time series DBs.

dominicchapman: In the past, I have said that event data is the record of every interaction between human, sensor, and machine. i can’t remember where that originated from, either cfrln or a deep dive on the history of event data ~Q1/Q2, but the reason I gravitated towards it is because it zooms us out from any use case first, to create an appreciation for the total scope and opportunity before honing in. 

By putting logs and traces as the first and second example, it feels to me that we have already coupled Axiom so tightly to engineering use cases, and then we try and walk it back with the third example, product interactions. I really want the reader to feel that timestamped event data is _the_ record of absolutely every digital interactional - the digital breadcrumb - and _then_ (and only then) for us to contextualize that with domain-specific examples. That could work better as two sentences.

dominicchapman: > While these can appear in various forms, from log lines to complex traces

Again, I feel the same way as above. This innocuous "from" scopes the breadth of use cases to engineering when I read this. The framing is changed from what I want the world to know, which is that every function in any business with digital activity can and will benefit from Axiom.

getting-started-guide/event-data.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+- **Timestamp**: When the event occurred.
+- **Attributes**: A set of key-value pairs offering details about the event context.
+- **Metadata**: Contextual labels and IDs, such as trace ID and span ID, that connect related events.
```
dominicchapman: For the avoidance of doubt, I feel the same way here about "such as trace ID and span ID".

getting-started-guide/event-data.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+Rather than simply calling them logs or metrics, event data reflects a broader range of interactions, crossing boundaries from engineering to product management, security, and beyond. This includes everything from the number of documents created in a product to the sequence of API calls during a checkout process.
```
dominicchapman: Good. 

"Whole engineers might most often work with timestamped event data in the form of logs or metrics..."

Something like this ^ might be more effective. I don’t love the phrasing, but the goal again would be to say: We know our most common reader is likely to be an engineer because of Axiom’s history, but that there is a bigger opp here for a business.

getting-started-guide/event-data.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+## Types of event data
```
dominicchapman: I know that you want to set up the punchline of Axiom’s metrics compatibility, but the problem here is we have now gone from what I hope will be a world view that timestamped event data shows up everywhere, to narrowing down even _more_ than engineering to observability.

If I look at this page, I would definitely leave thinking Axiom is just another engineering-focused tool. How could we change that perception? 

Does it need a heading that shows we are narrowing to the engineering use case specifically? Is this section or mention of observability needed at all?

getting-started-guide/event-data.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+Traditionally, observability has been associated with three pillars, each effectively a specialized view of event data:
+
+- **Logs**: Logs record discrete events, such as error messages or access requests, typically associated with engineering or security.
+- **Traces**: Traces track the path of requests through a system, capturing each step’s duration. By linking related spans within a trace, developers can identify bottlenecks and dependencies.
+- **Metrics**: Metrics quantify state over time, recording data like CPU usage or user count at intervals. Product or engineering teams can then monitor and aggregate these values for performance insights.
+
+In Axiom, these observability elements are stored as event data, allowing for fine-grained, efficient tracking across all three pillars.
```
dominicchapman: "Just like any other type of (timestamped) event data, ..."

getting-started-guide/event-data.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+## Metrics support
```
dominicchapman: Putting all of the above to one side, yes, this is good. It answers the question of our current metrics status, which is not well covered elsewhere.

dominicchapman (changes requested):
I want to see us start with the bigger story of timestamped event data, and then narrow to engineering consciously and clearly in the material. This reads the other way around to me: It starts with conventions and language from the engineering domain, and then tries to work up to a wider play. Based on our conversation, would you be open to trying one more revision? I think there is a great base here!

getting-started-guide/event-data.mdx
manototh: So far mistakenly considered not event data, but in fact also event data

manototh:
@dominicchapman I've implemented your suggestions. Most importantly, I've made the distinction clear between the general concept of event data and its particular use cases by splitting the original text into two pages: one about event data in general and its many use cases, and another about observability as a common (but not a singularly important) use case. Looking forward to what you think!

getting-started-guide/event-data.mdx
```diff
@@ -0,0 +1,29 @@
+---
+title: "Event data"
+description: "This page explains the fundamentals of timestamped event data in Axiom."
+sidebarTitle: Event data
+---
+
+Axioms’s mission is to help you leverage the power of timestamped event data.
```
dominicchapman:
```suggestion
Axiom’s mission is to operationalize every bit of event data in your organization.
```

Neil provided this specific definition of our mission earlier in the year, and I think it fits well here. The next paragraph then uses the longer form of timestamped event data, which feels ok to me.

getting-started-guide/event-data.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+Axiom considers timestamped event data to be the record of every digital interaction between human, sensor, and machine. Event data is the atomic unit of activity. For this reason, every function in any business with digital activity can benefit from leveraging event data.
```
dominicchapman:
```suggestion
Timestamped event data records every digital interaction between human, sensor, and machine, making it the atomic unit of activity for organizations. For this reason, every function in any business with digital activity can benefit from leveraging event data.
```

Do we need to scope this to Axiom’s belief system, or hold it as truth? This reads with more conviction to me.

getting-started-guide/event-data.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+Each event is simply a structured record—composed of key-value pairs—that captures meaningful interactions or changes in state within a system. While these can appear in various forms, they usually contain the following:
+
+- **Timestamp**: When the event occurred.
+- **Attributes**: A set of key-value pairs offering details about the event context.
+- **Metadata**: Contextual labels and IDs that connect related events.
```
dominicchapman, line 15: I’m not totally sure there is a distinction that needs to be drawn between attributes and metadata; I would probably use them interchangeably, but it doesn’t offend me.

getting-started-guide/event-data.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+<Note>
+The timestamped event data that Axiom works with is not to be confused with time-series event data.
```
dominicchapman: Without more context, I would omit this note. It doesn’t help the reader much imo. 

Usually, when using this kind of language, we’re creating a distinction between time series databases (like Timescale), and Axiom. 

I might draw on Neil’s idea that Axiom is suited to answering questions across operational (now, individual), analytical (recent history, population) and archival workloads, where time series databases might not have quite the same applied nature. But others at Axiom might be able to add more.

getting-started-guide/event-data.mdx
```diff
@@ -0,0 +1,29 @@
...
+</Note>
+
+## Uses of event data
+
+Event data, understood as the atomic unit of digital activity, is the lifeblood of modern businesses. Leveraging the power of event data is essential in the following areas, among others:
```
dominicchapman, line 19: Ok, this is a neat way to hone in on some specific use cases. 👍

getting-started-guide/event-data.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+- [Observability](/getting-started-guide/observability)
+- Security
+- Business intelligence
+- AI and machine learning
+- Finance
```
dominicchapman: Can we remove Finance and put Product analytics above business intelligence? That feels like a more natural selection and sequencing today. Sorry, not sure I can do multi-line suggestions inline in GitHub.

getting-started-guide/observability.mdx
```diff
@@ -0,0 +1,54 @@
+---
+title: "Axiom for observability"
+description: "This page explains how Axiom helps you leverage timestamped event data for observability purposes."
+sidebarTitle: Observability
+---
+
+Axiom helps you leverage the power of timestamped event data. A common use case of event data is observability (o11y) in the field of software engineering. Observability means measuring the health and performance of apps and distributed systems. The goal of collecting this information is to monitor the status of apps and systems, improve their quality, and fix errors.
```
dominicchapman: A more precise definition of observability that is usually reference is one’s ability to explain what is happening on the inside of a system just from observing it on the outside, or the ability to understand the behavior of (software) systems based on their outputs (telemetry, i.e. event data).

getting-started-guide/observability.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+Traditionally, observability has been associated with three pillars, each effectively a specialized view of event data:
+
+- **Logs**: Logs record discrete events, such as error messages or access requests, typically associated with engineering or security.
+- **Traces**: Traces track the path of requests through a system, capturing each step’s duration. By linking related spans within a trace, developers can identify bottlenecks and dependencies.
+- **Metrics**: Metrics quantify state over time, recording data like CPU usage or user count at intervals. Product or engineering teams can then monitor and aggregate these values for performance insights.
+
+In Axiom, these observability elements are stored as event data, allowing for fine-grained, efficient tracking across all three pillars.
+
+## Metrics support
```
dominicchapman: Given the paragraph before, I wonder if we shouldn’t at least touch on Axiom’s coverage of logs and distributed traces. For example, we could point to other pages in the documentation that are most relevant for those kinds of event data? 

Otherwise, I jump immediately to the part we do _not_ fully support, with its limitations.

dominicchapman (approved):
I like what you have done here Mano, great job.

I will approve this, as my suggestions could be incorporated in a follow up PR. I do think at least some elements of my recommendations should be baked in.

PR #157, opened 2024-11-22
Fix string literals

manototh:


thecraftman:
it looks like the search doesn't work on preview? :https://axiom-mano-string-literals.mintlify.app/introduction

I tried to search for string literal in the search bar but i don't see it. 

Screenshot 2024-11-22 at 18 23 05

thecraftman (approved):

PR #163, opened 2024-11-29
Update map fields and regions

manototh:
This PR resolves the following issues:
- [Map fields](https://axiom.co/docs/apl/data-types/map-fields)
  - Make clear that customers can’t create map fields themselves.
  - Remove custom map fields section, add note component after first section.
- [Regions](https://axiom.co/docs/reference/regions)
    - EU region is still in intro page of API docs.
    - Change heading of section, make text shorter, link to general Regions page.


dominicchapman (approved):
Very good! 

I question whether Map fields should be living under Query reference, versus Documentation tab, but one for another day.

PR #171, opened 2024-12-12
Update link in Datadog page

manototh:


send-data/datadog.mdx
```diff
@@ -9,7 +9,7 @@ logoId: 'datadog'
 
 Sending data from Datadog to Axiom is a private preview feature available upon request.
 
-Please [contact Axiom](https://axiom.co/migrations/datadog) to learn more about sending data from Datadog to Axiom.
+Please [contact Axiom](https://axiom.co/contact) to learn more about sending data from Datadog to Axiom.
```
simula-r: I think we actually want to keep things relative here? So maybe just /contact. 

Can you try that and let me know if it works?

manototh: Thanks for this. The problem is that it would be relative to `/docs/` which is not want we want.

simula-r: Hrmm i tried in the inspector and it worked fine... im gonna test it real fast.

simula-r: Yeah seems to work correctly with /contact you might want to clarify with mintlify. Generally speaking because you are using / its using the basepath of the proxy. Ill approve this but id check with them.

simula-r (approved):

send-data/datadog.mdx
manototh:
```suggestion
Please [contact Axiom](/contact) to learn more about sending data from Datadog to Axiom.
```

PR #174, opened 2024-12-16
Add inner join

manototh:
Preview: https://axiom-mano-inner-join.mintlify.app/apl/tabular-operators/join-operator


manototh:
I write that "to compare two fields of different data types, use on `$left.id_string == tostring($right.trace_id_int)`." At least this is how I expected it to work. This doesn't work unfortunately. It's possible to convert a single value like on `$left.status == tostring(500)`, but I couldn't figure out how to convert values from a whole field from the right side of the join.

apl/tabular-operators/join-operator.mdx
```diff
@@ -0,0 +1,175 @@
...
+- `Conditions`: The conditions for matching rows. The conditions determine how Axiom matches rows from the `LeftDataset` with rows from the `RightDataset`.
+    - To join datasets on a field that has the same name in the two datasets, simply use the field name. For example, `on id`.
+    - To join datasets on a field that has different names in the two datasets, define the two field names in an equality expression such as `on $left.id == $right.trace_id`.
+    - The two sides of the equality expression must have the same data type.
+    - You can use expressions in the join conditions. For example, to compare two fields of different data types, use `on $left.id_string == tostring($right.trace_id_int)`.
+    - You can define multiple join conditions. To separate conditions, use commas (`,`). Don’t use `and`. For example, `on $left.id == $right.trace_id, $left.span == $right.span_id`.
+
+### Returns
+
+The `join` operator returns a new table containing rows that match the specified join condition. The columns from the left and right tables are included, prefixed with `$left.` and `$right.` to differentiate overlapping names.
```
ted-gould: The returned table does not have the names specified. It is just the names of the fields of both columns.

apl/tabular-operators/join-operator.mdx
```diff
@@ -0,0 +1,175 @@
...
+### Parameters
+
+- `LeftDataset`: The first dataset, also known as the outer table or the left side of the join. If you expect one of the datasets to contain consistently less data than the other, specify the smaller dataset as the left side of the join.
+- `RightDataset`: The second dataset, also known as the inner table or the right side of the join.
+- `KindOfJoin`: Optionally, the [kind of join](#kinds-of-join) to perform. The only supported value and the default is `inner`.
+- `Conditions`: The conditions for matching rows. The conditions determine how Axiom matches rows from the `LeftDataset` with rows from the `RightDataset`.
+    - To join datasets on a field that has the same name in the two datasets, simply use the field name. For example, `on id`.
+    - To join datasets on a field that has different names in the two datasets, define the two field names in an equality expression such as `on $left.id == $right.trace_id`.
+    - The two sides of the equality expression must have the same data type.
+    - You can use expressions in the join conditions. For example, to compare two fields of different data types, use `on $left.id_string == tostring($right.trace_id_int)`.
```
ted-gould: You can't use functions and `$left` and `$right`. It is only available if you want to specify fields. By default, left is the left side of the equals and right is the right. So you really only need them if you want to do `$right.right_id == $left.left_id`, which is diabolical :-)

apl/tabular-operators/join-operator.mdx
```diff
@@ -0,0 +1,175 @@
...
+```kusto
+LeftDataset
+| join kind=KindOfJoin RightDataset on Conditions
+```
+
+### Parameters
+
+- `LeftDataset`: The first dataset, also known as the outer table or the left side of the join. If you expect one of the datasets to contain consistently less data than the other, specify the smaller dataset as the left side of the join.
+- `RightDataset`: The second dataset, also known as the inner table or the right side of the join.
+- `KindOfJoin`: Optionally, the [kind of join](#kinds-of-join) to perform. The only supported value and the default is `inner`.
```
ted-gould: Probably can leave off supported here, worried we'll forget to update it :-) Let them click the link to learn more there.

apl/tabular-operators/join-operator.mdx
```diff
@@ -110,19 +110,19 @@ LeftDataset
...
+- `RightDataset`: The second dataset, also known as the inner dataset or the right side of the join.
+- `KindOfJoin`: Optionally, the [kind of join](#kinds-of-join) to perform.
 - `Conditions`: The conditions for matching rows. The conditions determine how Axiom matches rows from the `LeftDataset` with rows from the `RightDataset`.
     - To join datasets on a field that has the same name in the two datasets, simply use the field name. For example, `on id`.
     - To join datasets on a field that has different names in the two datasets, define the two field names in an equality expression such as `on $left.id == $right.trace_id`.
     - The two sides of the equality expression must have the same data type.
-    - You can use expressions in the join conditions. For example, to compare two fields of different data types, use `on $left.id_string == tostring($right.trace_id_int)`.
+    - You can use expressions in the join conditions. For example, `on $left.status_string == tostring(500)`.
```
ted-gould: Probably shouldn't have this be a constant. So it would be better to be `on $left.status_string == tostring(trace_id_int)` where `trace_id_int` is in the right dataset.

ted-gould (approved):

PR #175, opened 2024-12-16
Add monitor status page

manototh:
Preview: https://axiom-mano-monitor-status.mintlify.app/monitor-data/configure-monitors

### DIFF
--- a/monitor-data/configure-monitors.mdx
+++ b/monitor-data/configure-monitors.mdx
@@ -6,6 +6,38 @@ tags:
     ['axiom documentation', 'documentation', 'axiom', 'dashboards', 'monitors', 'notifiers', 'slack', 'email', 'pagerduty', 'webhook', 'opsgenie', 'discord', 'discord webhook']
 ---
 
+## View monitor status
+
+To view the status of a monitor:
+1. Click the Monitors tab.
+1. Click the monitor in the list whose status you want to view.
+
+The monitor status page provides an overview of the current status and the history of the monitor.
+
+On the left, you see the recent activity and the history of the monitor runs:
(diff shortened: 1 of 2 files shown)


monitor-data/configure-monitors.mdx
```diff
@@ -6,6 +6,38 @@ tags:
...
+- The `Status` field displays the status of the monitor.
+- The `Range from` and `Range to` fields display the time range used in the monitor run.
+
+You can change the time range of this overview in the top right corner.
+
+On the right, you see details about the monitor’s configuration.
+- Current status
+- Monitor type
+- Query the monitor periodically runs
+- Configuration details
```
benenharrington, line 28: This reads funny to me, since it appears in a list of things that we've referred to as 'details about the monitor's configuration'

monitor-data/configure-monitors.mdx
```diff
@@ -6,6 +6,38 @@ tags:
...
+- Notifiers attached to the monitor
+- Metadata such as name and description
+
+## Change monitors
```
benenharrington, line 32: I'm guessing this is deliberate, but why isn't this 'Edit'?

monitor-data/configure-monitors.mdx
manototh: Thanks! Corrected.

monitor-data/configure-monitors.mdx
manototh: We tend to use `change` or `update` in headings. `Edit` is the label of the button. I think these are all synonyms so it's just a matter of stylistic choice.

benenharrington (approved):

PR #176, opened 2024-12-16
Add timestamp-based pagination

manototh:
Preview: https://axiom-mano-time-based-pagination.mintlify.app/restapi/pagination#timestamp-based-pagination


restapi/pagination.mdx
```diff
@@ -12,25 +12,134 @@ You can use pagination for the following endpoints:
...
+- Include the [`limit` operator](/apl/tabular-operators/limit-operator) in the APL query of your API request. The argument of this operator determines the number of events to display per page.
+- Use `sort by _time asc` or `sort by _time desc` in the APL query. This returns the results in ascending or descending chronological order. For more information, see [sort operator](/apl/tabular-operators/sort-operator).
+- Specify `startTime` and `endTime` in the body of your API request.
+
+### Run Query (Legacy)
+
+To use timestamp-based pagination with the legacy Run Query endpoint:
+- Add the `limit` parameter to the body of your API request. The value of this parameter determines the number of events to display per page.
+- Add the `order` parameter to the body of your API request. In the value of this parameter, order the results by time in either ascending or descending chronological order. For example, `[{ "field": "_time", "desc": true }]`. For more information, see [order operator](/apl/tabular-operators/order-operator).
+- Specify `startTime` and `startTime` in the body of your API request.
```
darach:
```suggestion
- Specify `startTime` and `endTime` in the body of your API request.
```

darach: Fix typo

restapi/pagination.mdx
```diff
@@ -12,25 +12,134 @@ You can use pagination for the following endpoints:
...
+## Page through the result set
+
+Use the timestamps as boundaries to page through the result set.
+
+### Queries with descending order
+
+To go to the next page of the result set for queries with descending order (`_time desc`):
+1. Determine the timestamp of last item on the current page. This is the least recent event.
+1. Subtract 1 nanosecond from the timestamp.
+1. In your next request, change the value `endTime` parameter in the body of your API request to the timestamp of the last item minus 1 nanosecond.
```
darach: We've used user defined pagination with and without the nanosecond increments
and found it slightly better without. There is a risk of duplication in one, and overlap
in the other. Perhaps suggest that regardless of the method that some deduplication
may be required.

Same goes for ascending/descending order. Duplicates are possible for many reasons
including backfill or natural duplication from external data sources so it is something
a pagination script would ordinarily need to consider...

restapi/pagination.mdx
```diff
@@ -12,25 +12,134 @@ You can use pagination for the following endpoints:
...
+
+To use timestamp-based pagination with the Run Query endpoint:
+- Include the [`limit` operator](/apl/tabular-operators/limit-operator) in the APL query of your API request. The argument of this operator determines the number of events to display per page.
+- Use `sort by _time asc` or `sort by _time desc` in the APL query. This returns the results in ascending or descending chronological order. For more information, see [sort operator](/apl/tabular-operators/sort-operator).
+- Specify `startTime` and `endTime` in the body of your API request.
+
+### Run Query (Legacy)
+
+To use timestamp-based pagination with the legacy Run Query endpoint:
+- Add the `limit` parameter to the body of your API request. The value of this parameter determines the number of events to display per page.
```
darach, line 38: The legacy and tabular APIs allow request based `limit` configuration. This means that the minimum or lesser of the query limit and request limit will be in force up to the maximum internal supported limit that axiom supports server side.

In practice, Axiom will default the limit to some value - currently 1000 events. For pagination of datasets that are greater than 1000 events we've found it best to pin the limit in the request and in the APL query to the same number to avoid the defaults and any confusion w.r.t `min(request_limit, apl_limit, axiom_default_limit, max_axiom_limit)` ...

darach (changes requested):
I left a few clarifications / context and one suggested fix. The suggested fix is a typo, the clarifications are optional / just suggested

darach (approved):
LGTM 👍

PR #186, opened 2025-01-15
Add information about specifying log levels

manototh:
Preview: https://axiom-mano-log-levels.mintlify.app/reference/field-restrictions#requirements-for-log-level-fields
https://axiom-mano-log-levels.mintlify.app/query-data/stream#highlight-severity


reference/field-restrictions.mdx
```diff
@@ -63,6 +63,35 @@ If you include the `_time` field in the ingested data, follow these requirements
...
+- `@severity`
+- `status.code`
+
+Axioms recommends specifying the log level in one of the explicit status fields mentioned above. If Axiom doesn’t detect the log level in one of these fields, it searches further in the following fields:
+- `message`
+- `@message`
+- `msg`
+- `record`
+- `text`
+
```
flbn: tbh these are our best guesses and i would recommend _not_ pointing that out, for the sake of it being confusing already

flbn (changes requested):
other than the blurb i pointed out this is good, i would have @benenharrington look over it though. i think team laharadan could potentially have some reservations, they were going to look into reworking our approach to log levels at some point?

reference/field-restrictions.mdx
manototh: Sure, removed

flbn (approved):

PR #191, opened 2025-01-22
Add Security and Product analytics use case to Learn Axiom section

manototh:
https://axiom-mano-security-product-analytics.mintlify.app/getting-started-guide/product-analytics
https://axiom-mano-security-product-analytics.mintlify.app/getting-started-guide/security


dominicchapman:
We talked in 1-1 about opportunity to augment this conceptual base with applied examples. Capturing some of that here:

In the case of product analytics, how might a B2B SaaS company add instrumentation to their apps and services, ship relevant events to Axiom, and then query or even monitor that event data? Some examples might include Account (our "organization") Created, (user) Signed Up, and Trial Started. Feels like there is a good story to tell on operational workflows (what is happening right now in our product? e.g. a user just upgraded to a paid plan) and analytical workloads (what happened in aggregate in the last 3 months)?

A separate point: Terms like groundbreaking feel overbearing, artificial or boastful. Might be a personal taste thing.

As a base, I’m excited about where this material is going to set the stage for the wider set of use cases.

PR #195, opened 2025-01-23
docs: Add docs for axiom-logging

gabrielelpidio:


send-data/javascript-logging.mdx
```diff
@@ -0,0 +1,172 @@
...
+`AxiomFetchTransport`: Sends logs to Axiom using the Axiom HTTP API.
+
+```ts
+import { AxiomFetchTransport } from "@axiomhq/logging";
+
+const transport = new AxiomFetchTransport({
+  token: process.env.AXIOM_TOKEN,
+  dataset: process.env.AXIOM_DATASET,
+  url: process.env.AXIOM_URL,
+  autoFlush: 1000,
```
islameldigwi: this `autoFlush` with 1000 I guess needs to be more clear on what it does, is that 1000ms or 1000 log events? maybe it needs to an object instead with clear naming.

send-data/javascript-logging.mdx
```diff
@@ -0,0 +1,172 @@
...
+});
+
+const transport = new AxiomJSTransport({
+  axiom,
+  dataset: process.env.AXIOM_DATASET,
+  logLevel: "warn",
+});
+```
+
+- `ProxyTransport`: Sends logs to a remote server that acts as a proxy.
```
islameldigwi: This proxy is mean for client side to use a route handler as proxy to Axiom, right? Needs better description and the example should just be about using an internal api endpoint.

send-data/react.mdx
```diff
@@ -0,0 +1,95 @@
...
+  ]
+isPopular: true
+popularityOrder: 9
+logoId: "react"
+---
+
+## Installation
+
+```sh
+npm install --save @axiomhq/react @axiomhq/logging
```
islameldigwi: lets make `@axiomhq/logging` come before the react package, to make it clear that this is as a dependency for the react lib.

send-data/javascript-logging.mdx
islameldigwi, line 1: While the doc is great, I feel like it serves as a ref instead of a usage guide. I would prefer if:
- The top section describes the easiest use case a long with one transport
- Then a section on formatting events. Afterwards, listing different types of transports
- Perhaps we should have a note that links the user to react and next.js packages
- @tothmano would a table of content on top be useful here?

thesollyz (changes requested):

manototh:
Closing in favour of https://github.com/axiomhq/docs/pull/215

PR #198, opened 2025-01-28
Add redact

manototh:
Preview: https://axiom-mano-redact.mintlify.app/apl/tabular-operators/redact-operator

Context: https://github.com/gordallott/axiom/blob/62d7394faa7cbfebae31c21d4a6de60df44210fc/pkg/kirby/apl/parser/ast/v2/ops.go#L1438-L1456


apl/tabular-operators/redact-operator.mdx
```diff
@@ -0,0 +1,150 @@
...
+
+```kusto
+| redact [replaceToken="*"] [replaceHash=false] [redactGroups=false] <regex>, (<regex>) [on Field]
+```
+
+### Parameters
+
+| Parameter       | Type     | Description                                                                 |
+|-----------------|----------|-----------------------------------------------------------------------------|
+| `replaceToken`  | string   | The character with which to replace matches. The default is `*`.                     |
```
gordallott: There is a slight quirk here, in that if the replaceToken is a single character, we replace all the characters that match, like:
```
| extend foo = "foo bar baz"
| redact replaceToken="x" "foo" on foo 
```
 => `"xxx bar baz"` 

but if the replaceToken is more than one character then we replace the whole matching string with that token, for example
```
| extend foo = "foo bar baz"
| redact replaceToken="abc" "foo" on foo
```
=> `"abc bar baz"` 

which can be useful if you want to indicate that something was redacted, for example
```
| extend foo = "foo bar baz"
| redact replaceToken="<REDACTED>" "bar" on foo 
```
=> `"foo <REDACTED> baz"`

gordallott (approved):
just a clarifying comment otherwise all good :+1:

apl/tabular-operators/redact-operator.mdx
manototh: @gordallott Many thanks for this. Great point. Implemented!

PR #212, opened 2025-02-11
Correct the description of anomaly monitor tolerance

benenharrington:
Current description of tolerance implies its a fixed value added/subtracted from the expected value to determine the non-alerting range. This PR clarifies that the tolerance is combined with a measure of the amount of variation in the monitor query to determine the range.


monitor-data/anomaly-monitors.mdx
```diff
@@ -17,7 +17,7 @@ To create an anomaly monitor, follow these steps:
...
     - The comparison operator is the rule to apply when comparing the results to the expected value. The possible values are **above**, **below**, and **above or below**.
-    - The tolerance factor controls the sensitivity of the monitor. It defines how much deviation from the expected value to accept without triggering the monitor. It’s a range above and below the expected value. The higher the tolerance factor, the wider this range. When the results of the aggregation stay within this range, the monitor doesn’t trigger. When the results of the aggregation cross this range, the monitor triggers. The tolerance factor can be any positive numeric value.
+    - The tolerance factor controls the sensitivity of the monitor. It combines with a measure of how much the results of your query tend to vary to determine how much deviation from the expected value is required to trigger the monitor. The higher the tolerance factor, the wider this range. When the results of the aggregation stay within this range, the monitor doesn’t trigger. When the results of the aggregation cross this range, the monitor triggers. The tolerance factor can be any positive numeric value.
```
manototh:
```suggestion
    - The tolerance factor controls the sensitivity of the monitor. Axiom combines the tolerance factor with a measure of how much the results of your query tend to vary, and uses them to determine how much deviation from the expected value to tolerate before triggering the monitor. The higher the tolerance factor, the wider the tolerated range of deviation. When the results of the aggregation stay within this range, the monitor doesn’t trigger. When the results of the aggregation cross this range, the monitor triggers. The tolerance factor can be any positive numeric value.
```

manototh: @benenharrington Thanks for this. What do you think?

benenharrington: Thanks! 👍

manototh (approved):

PR #213, opened 2025-02-11
Update syslog-proxy.mdx

darach:
Update install to use go install instead of go get and bias for recent golang editions.
Spotted by an eagle eyed customer.


manototh (approved):

manototh:
@darach Thanks for this!

PR #215, opened 2025-02-12
Add new axiom-js libraries (logging, react, nexjs)

manototh:
Preview:
https://axiom-mano-axiom-js-repos.mintlify.app/send-data/nextjs
https://axiom-mano-axiom-js-repos.mintlify.app/guides/javascript
https://axiom-mano-axiom-js-repos.mintlify.app/send-data/react


guides/javascript.mdx
```diff
@@ -95,6 +129,160 @@ await annotations.create({
...
+    });
+
+    const transport = new AxiomJSTransport({
+      axiom,
+      dataset: process.env.AXIOM_DATASET,
+      logLevel: "warn",
+    });
+    ```
+
+- `ProxyTransport`: Sends logs to a remote server that acts as a proxy.
```
manototh: From @dasfmi 
This proxy is mean for client side to use a route handler as proxy to Axiom, right? Needs better description and the example should just be about using an internal api endpoint.

guides/javascript.mdx
```diff
@@ -95,6 +129,160 @@ await annotations.create({
...
+- `AxiomFetchTransport`: Sends logs to Axiom using the Axiom HTTP API.
+
+    ```ts
+    import { AxiomFetchTransport } from "@axiomhq/logging";
+
+    const transport = new AxiomFetchTransport({
+      token: process.env.AXIOM_TOKEN,
+      dataset: process.env.AXIOM_DATASET,
+      url: process.env.AXIOM_URL,
+      autoFlush: 1000,
```
manototh: From @dasfmi 
this autoFlush with 1000 I guess needs to be more clear on what it does, is that 1000ms or 1000 log events? maybe it needs to an object instead with clear naming.

manototh:
From @dasfmi

While the doc is great, I feel like it serves as a ref instead of a usage guide. I would prefer if:

The top section describes the easiest use case a long with one transport
Then a section on formatting events. Afterwards, listing different types of transports
Perhaps we should have a note that links the user to react and next.js packages
@tothmano would a table of content on top be useful here?

manototh:
@dasfmi Many thanks for your comments. I agree that the current page on JS is a reference for advanced users rather than a step-by-step guide that anyone can follow. I think it will work for the initial release and I will make time for restructuring it later.

manototh:
@gabrielelpidio Many thanks for the initial draft. I've tried to my best to formulate it in a way that users know what to choose from this wide array of options. Please let me know what you think.

gabrielelpidio (approved):

PR #216, opened 2025-02-12
Change Optimize performance page to troubleshooting resource

manototh:
Preview: https://axiom-mano-performance.mintlify.app/reference/performance#practices-that-limit-axiom-performance
Source: https://www.notion.so/axiomhq/Practices-that-limit-Axiom-performance-and-how-to-avoid-them-197fc65a3019802e99e4e0b0ee36791f

### DIFF
--- a/reference/performance.mdx
+++ b/reference/performance.mdx
@@ -80,7 +80,7 @@ These issues reduce compression efficiency, make indexing less effective, and fo
 
 - **Keep datasets narrowly focused:** Group data from the same application or service in its own dataset. For example, keep `k8s_logs` separate from `web_traffic`.
 - **Avoid mixing data types for the same field:** Enforce consistent types during ingest. If a field is numeric, always send numeric values.
-- **Consider using map fields:** If you have sparse or high-cardinality nested data, consider storing it in a single map (object) field instead of flattening every key. This reduces the total number of top-level fields. Axiom’s [map fields](https://axiom.co/docs/apl/data-types/map-fields#map-fields) are optimized for large objects.
+- **Consider using map fields:** If you have sparse or high-cardinality nested data, consider storing it in a single map (object) field instead of flattening every key. This reduces the total number of top-level fields. Axiom’s [map fields](/apl/data-types/map-fields#map-fields) are optimized for large objects.
 
 ### Large time ranges mixed in the same ingest
 
@@ -154,7 +154,7 @@ By default, Axiom’s query engine projects all fields (`project *`) for each ma
     
     Specify exactly which fields you need. For example:
     
-    ```
+    ```kusto
     dataset
     | where status == 500
     | project timestamp, error_code, user_id
@@ -162,7 +162,7 @@ By default, Axiom’s query engine projects all fields (`project *`) for each ma
     
 - **Use `project-away` if you only need to exclude a few fields:** If you need 90% of the fields but want to exclude the largest ones, for instance:
     
-    ```
+    ```kusto
     dataset
     | project-away debug_payload, large_object_field
     
@@ -190,12 +190,14 @@ Regular expressions (`matches`, `regex`) can be powerful, but they are also expe
     
     Instead of:
     
-    ```
+    ```kusto
     dataset
     | where message matches "[Ff]ailed"
     ```
+
+    Use:
     
-    ```
+    ```kusto
     dataset
     | where message contains "failed"
     ```
@@ -204,14 +206,14 @@ Regular expressions (`matches`, `regex`) can be powerful, but they are also expe
     
     If you simply need to find “foobar” anywhere in the text, use:
     
-    ```
+    ```kusto
     dataset
     | search "foobar"
     ```
     
-    or:
+    Alternatively:
     
-    ```
+    ```kusto
     dataset
     | where FIELD contains_cs "foobar"
     ```
@@ -232,11 +234,11 @@ Some ingestion pipelines place large JSON payloads into a string field, deferrin
 
 #### How to fix it
 
-- **Ingest as map fields:** Axiom’s new [map column type](https://axiom.co/docs/apl/data-types/map-fields#map-fields) can store object fields column by column, preserving structure and optimizing for nested queries. This allows indexing of specific nested keys.
+- **Ingest as map fields:** Axiom’s new [map column type](/apl/data-types/map-fields#map-fields) can store object fields column by column, preserving structure and optimizing for nested queries. This allows indexing of specific nested keys.
 - **Extract top-level fields where possible:** If a certain nested field is frequently used for filtering or grouping, consider promoting it to its own top-level column (for faster indexing).
 - **Avoid `parse_json()` in query:** If your JSON cannot be flattened entirely, ingest it into a map field. Then query subfields directly:
     
-    ```
+    ```kusto
     dataset
     | where data_map.someKey == "someValue"
     ```
@@ -258,18 +260,16 @@ You can create virtual fields (for example, `extend converted = toint(some_field
     
     **Example:** Instead of
     
-    ```
+    ```kusto
     dataset
     | extend str_user_id = tostring(mixed_user_id)
     | where str_user_id contains "123"
     ```
     
-    simply do
+    Use:
     
-    ```
-    
+    ```kusto
     | where mixed_user_id contains "123"
-    
     ```
     
     The filter automatically matches string values in mixed columns.
@@ -313,7 +313,7 @@ Axiom’s query engine does not currently reorder your `where` clauses optimally
     
     Example:
     
-    ```
+    ```kusto
     dataset
     | where user_id == 1234
     | where log_level == "ERROR"
--- a/reference/performance.mdx
+++ b/reference/performance.mdx
@@ -12,12 +12,11 @@ Axiom is optimized for storing and querying timestamped event data. However, cer
 | Practice | Severity | Impact |
 | --- | --- | --- |
 | [Mixing unrelated data in datasets](#mixing-unrelated-data-in-datasets) | Critical | Combining unrelated data inflates schema, slows queries |
-| [Large time ranges in one ingest](#large-time-ranges-mixed-in-the-same-ingest) | Critical | Breaks time-based indexing |
+| [Excessive backfilling, big difference between _time and _sysTime](#excessive-backfilling-and-large-_time-vs-_systime-gaps) | Critical | Creates overlapping blocks, breaks time-based indexing |
 | [Large number of fields in a dataset](#large-number-of-fields-in-a-dataset) | High | Very high dimensionality slows down query performance |
 | [Failing to use _time](#failing-to-use-the-_time-field-for-event-timestamps) | High | No efficient time-based filtering |
 | [Overly wide queries (project *)](#overly-wide-queries-returning-more-fields-than-needed) | High | Returns massive unneeded data |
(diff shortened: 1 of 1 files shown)


manototh:
@peterbourgon @dominicchapman @tsenart Based on Peter's [feedback](https://www.notion.so/axiomhq/Practices-that-limit-Axiom-performance-and-how-to-avoid-them-197fc65a3019802e99e4e0b0ee36791f), I've merged the section on  `Large time ranges` into `Excessive backfilling`: https://github.com/axiomhq/docs/pull/216/commits/0683d84738a296bcf40c93e503d677756f387d9c

Please let me know if this work. Thanks!

reference/performance.mdx
```diff
@@ -5,36 +5,288 @@ sidebarTitle: Performance
...
+
+### Problem
+
+A “kitchen-sink” dataset is one in which events from multiple, unrelated applications or services get lumped together, often resulting in:
+
+- **Excessive width (too many columns)**: Adding more and more unique fields bloats the schema, reducing query throughput.
+- **Mixed data types in the same field**: For example, some events store `user_id` as a string, while others store it as a number in the same `user_id` field.
+- **Unrelated schemas in a single dataset**: Columns that make sense for one app might be `null` or typed differently for another.
+
+These issues reduce compression efficiency, make indexing less effective, and force Axiom to scan more data than necessary.
```
tsenart: This doesn't affect `_time` indexing effectiveness.

```suggestion
These issues reduce compression efficiency and force Axiom to scan more data than necessary.
```

reference/performance.mdx
```diff
@@ -5,36 +5,288 @@ sidebarTitle: Performance
...
+
+### Why it matters
+
+- **Slower queries**: Each query must parse wider blocks of data and handle inconsistent column types.
```
tsenart:
```suggestion
- **Slower queries**: Each query must scan wider blocks of data and handle inconsistent column types.
```

reference/performance.mdx
```diff
@@ -5,36 +5,288 @@ sidebarTitle: Performance
...
+- **Higher resource usage**: Wide schemas reduce row packing in blocks, harming throughput and potentially raising costs.
+- **Harder data exploration**: When fields differ drastically between events, discovering the correct columns or shaping queries becomes more difficult.
+
+### How to fix it
+
+- **Keep datasets narrowly focused:** Group data from the same application or service in its own dataset. For example, keep `k8s_logs` separate from `web_traffic`.
```
tsenart, line 45: This is a tricky example, because `k8s_logs` typically, and by definition, receive logs from all the services running on top of a kubernetes cluster. I'd suggest that for datasets where query performance is critical, splitting it off of a kitchen sink dataset is ideal, but starting off with a dataset per application is too onerous an ask, and unnecessary for low to mid scale.

reference/performance.mdx
```diff
@@ -5,36 +5,288 @@ sidebarTitle: Performance
...
+- **Avoid mixing data types for the same field:** Enforce consistent types during ingest. If a field is numeric, always send numeric values.
+- **Consider using map fields:** If you have sparse or high-cardinality nested data, consider storing it in a single map (object) field instead of flattening every key. This reduces the total number of top-level fields. Axiom’s [map fields](/apl/data-types/map-fields#map-fields) are optimized for large objects.
+
+## Excessive backfilling and large `_time` vs. `_sysTime` gaps
+
+### Problem
+
+Axiom’s index is primarily built around time. Ideally, incoming events for a block lie in a closely bounded time range. However, backfilling large amounts of historical data after the fact (especially out of chronological order) creates wide time overlaps in blocks. If `_time` is far from `_sysTime` (the time the event was ingested), Axiom’s primary time index is weakened.
```
tsenart: Removing ambiguity. We don't have other indexes at the moment. So, "primary" doesn't make sense.

```suggestion
Axiom’s `_time` index is critical for query performance. Ideally, incoming events for a block lie in a closely bounded time range. However, backfilling large amounts of historical data after the fact (especially out of chronological order) creates wide time overlaps in blocks. If `_time` is far from `_sysTime` (the time the event was ingested), Axiom’s time index effectiveness is weakened.
```

reference/performance.mdx
```diff
@@ -5,36 +5,288 @@ sidebarTitle: Performance
...
+
+### Why it matters
+
+- **Poor performance on time-based queries**: Blocks must be scanned despite time filters, because many blocks overlap the query time window.
+- **Inefficient block filtering**: Queries that filter on time must scan blocks that contain data from many time segments.
```
tsenart:
```suggestion
- **Inefficient block filtering**: Queries that filter on time must scan blocks that contain data from a wide time range.
```

reference/performance.mdx
```diff
@@ -5,36 +5,288 @@ sidebarTitle: Performance
...
+## Large number of fields in a dataset
+
+### Problem
+
+Slow query performance in datasets with very high dimensionality (with more than several thousand fields).
+
+### Why it matters
 
 Axiom stores event data in a tuned format. As a result:
 - The number of distinct values (cardinality) in your data has little impact on performance.
```
tsenart: We may not want to make this point here, but technically, low cardinality fields compress better than high cardinality fields, so it does impact performance in the sense that we achieve higher compression ratios than otherwise.

reference/performance.mdx
```diff
@@ -5,36 +5,288 @@ sidebarTitle: Performance
...
+### How to fix it
+
 Scoping the number of fields in a dataset below a few thousand can help you achieve the best performance in Axiom.
 
-## Optimize queries
+## Failing to use the `_time` field for event timestamps
+
+### Problem
+
+Axiom’s core optimizations rely on `_time` for indexing and time-based queries. If you store event timestamps in a different field (for example, `timestamp` or `created_at`) and use that field in time filters, Axiom’s time-based optimizations will not be applied.
```
tsenart:
```suggestion
Axiom’s core optimizations rely on `_time` for indexing and time-based queries. If you store event timestamps in a different field (for example, `timestamp` or `created_at`) and use that field in time filters, Axiom’s time-based optimizations will not be leveraged.
```

reference/performance.mdx
```diff
@@ -5,36 +5,288 @@ sidebarTitle: Performance
...
+
+### Why it matters
+
+- **No time-based indexing**: Every block must be scanned, because your custom timestamp field is invisible to the time index.
+- **Incorrect event ordering**: You may lose correct chronological context or get unexpected query results.
```
tsenart: This is inaccurate. We do maintain correct sorting order on any field we explicitly sort on, it just won't be efficient.

reference/performance.mdx
```diff
@@ -5,36 +5,288 @@ sidebarTitle: Performance
...
+
+### How to fix it
+
+- **Always use `_time`:** Configure your ingest pipelines so that Axiom sets `_time` to the actual event timestamp.
+    - If you have a custom field like `created_at`, map it to `_time` at ingest.
```
tsenart: Not only map it, but rename it, so that we don't keep around two fields with identical data, which is bad for compression.

reference/performance.mdx
```diff
@@ -5,36 +5,288 @@ sidebarTitle: Performance
...
+
+- **Standardize your types at ingest:** If a field is semantically an integer, always send it as an integer.
+- **Use consistent schemas across services:** If multiple applications write to the same dataset, agree on a schema and data types.
+- **Perform corrections at the source:** If you discover your data has been mixed historically, stop ingesting mismatched types. Over time, new blocks will reflect the corrected types even though historical blocks remain mixed.
+
+## Overly wide queries returning more fields than needed
+
+### Problem
+
+By default, Axiom’s query engine projects all fields (`project *`) for each matching event. This can return large amounts of unneeded data, especially in wide datasets with many columns.
```
tsenart:
```suggestion
By default, Axiom’s query engine projects all fields (`project *`) for each matching event. This can return large amounts of unneeded data, especially in wide datasets with many fields.
```

reference/performance.mdx
```diff
@@ -5,36 +5,288 @@ sidebarTitle: Performance
...
+    
+    ```kusto
+    dataset
+    | project-away debug_payload, large_object_field
+    
+    ```
+    
+- **Limit your results**
+    
+    If you only need a sample of events for debugging, use a lower `limit` value (such as 100) instead of the default 1000.
```
tsenart:
```suggestion
    If you only need a sample of events for debugging, use a lower `limit` value (such as 10) instead of the default 1000.
```

reference/performance.mdx
```diff
@@ -5,36 +5,288 @@ sidebarTitle: Performance
...
+    | where message contains "failed"
+    ```
+    
+- **Use `search` for substring search:**
+    
+    If you simply need to find “foobar” anywhere in the text, use:
+    
+    ```kusto
+    dataset
+    | search "foobar"
```
tsenart, line 195: Search is far more expensive than using a regex on a single field, because it needs to match *all* fields.

reference/performance.mdx
```diff
@@ -5,36 +5,288 @@ sidebarTitle: Performance
...
+
+### Why it matters
+
+- **Repeated parsing overhead**: You pay a performance penalty on each query.
+- **Limited indexing**: Axiom cannot index nested fields if they are only known at query time.
+
+### How to fix it
+
+- **Ingest as map fields:** Axiom’s new [map column type](/apl/data-types/map-fields#map-fields) can store object fields column by column, preserving structure and optimizing for nested queries. This allows indexing of specific nested keys.
+- **Extract top-level fields where possible:** If a certain nested field is frequently used for filtering or grouping, consider promoting it to its own top-level column (for faster indexing).
```
tsenart:
```suggestion
- **Extract top-level fields where possible:** If a certain nested field is frequently used for filtering or grouping, consider promoting it to its own top-level column (for faster scanning and filtering).
```

reference/performance.mdx
```diff
@@ -5,36 +5,288 @@ sidebarTitle: Performance
...
+    ```kusto
+    dataset
+    | where data_map.someKey == "someValue"
+    ```
+
+## Virtual fields for simple transformations
+
+### Problem
+
+You can create virtual fields (for example, `extend converted = toint(some_field)`) to transform data at query time. While sometimes necessary, every additional virtual field imposes more overhead.
```
tsenart:
```suggestion
You can create virtual fields (for example, `extend converted = toint(some_field)`) to transform data at query time. While sometimes necessary, every additional virtual field imposes overhead.
```

tsenart (approved):

manototh:
@tsenart Many thanks for this. I have implemented your suggestions here: https://github.com/axiomhq/docs/pull/216/commits/41ad190fed297b70385597325979ce542757382c
Could you please take a final look to be safe? Thanks in advance.

reference/performance.mdx
```diff
@@ -95,13 +95,13 @@ Axiom’s core optimizations rely on `_time` for indexing and time-based queries
 
 ### Why it matters
 
-- **No time-based indexing**: Every block must be scanned, because your custom timestamp field is invisible to the time index.
-- **Incorrect event ordering**: You may lose correct chronological context or get unexpected query results.
+- **No time-based indexing**: Every block must be scanned because your custom timestamp field is invisible to the time index.
+- **Inefficient event ordering**: Queries that sort on time are slow.
```
tsenart: Queries that sort on `_time` are not slow. Queries that sort on any other field for which we have no index are slower, but still correct.

PR #220, opened 2025-02-19
Ensure sharing page shows in sidebar

benenharrington:
- Updates sidebar to ensure sharing docs discoverable
- Adds caveat around link deletion not removing shared access


darach (approved):
LGTM 👍

PR #241, opened 2025-03-06
Add search within query results

manototh:
Preview: https://axiom-mano-search.mintlify.app/query-data/explore#search-within-query-results


query-data/explore.mdx
```diff
@@ -246,4 +246,14 @@ In the event timeline, line charts, and heat maps, you can drag the pointer over
...
+
+### Search within query results
+
+To quickly search for an expression and highlight its occurrences within the query results:
+
+1. In the query results view, press **CMD+F** (Mac) / **CTRL+F** (Windows).
+1. Type the expression that you want to search for.
+1. Press **Enter**.
+
+Currently, the search only highlights matching entries in the events rendered on your screen.
```
flbn: we merged column view late last night!

query-data/explore.mdx
```diff
@@ -246,4 +246,14 @@ In the event timeline, line charts, and heat maps, you can drag the pointer over
...
 
-The time range of your query automatically updates to match what you selected.
\ No newline at end of file
+The time range of your query automatically updates to match what you selected.
+
+### Search within query results
+
+To quickly search for an expression and highlight its occurrences within the query results:
+
+1. In the query results view, press **CMD+F** (Mac) / **CTRL+F** (Windows).
```
dominicchapman: Thoughts on adopting [this](https://linear.app/docs/editor) kind of style e.g. `Cmd/Ctrl+F` for these kinds of shortcuts? I think our audience will be comfortable knowing what the context where each is appropriate so it’s a simpler shorthand

query-data/explore.mdx
dominicchapman: It’s the opposite of this. Using browser search would be limited to highlighting matching entries in events rendered on screen since the list is virtualized, but Axiom’s `Cmd/Ctrl+F` search feature will return every relevant search response in the entire virtualized list (i.e. it will match entries from all results returned by the query - respecting automatic limits like 1,000).

dominicchapman (dismissed):
The part about matching entries on rendered screen is misleading AFAICT, so would should update that my understanding is accurate - otherwise, great!

flbn (approved):

query-data/explore.mdx
tothmano: Good idea, fixed, thanks! Also HTML styling work for keyboard buttons.
image

query-data/explore.mdx
manototh: Good idea, fixed, thanks! Also HTML styling work for keyboard buttons.
image

query-data/explore.mdx
```diff
@@ -246,4 +246,15 @@ In the event timeline, line charts, and heat maps, you can drag the pointer over
...
+### Search within query results
+
+To quickly search for an expression and highlight its occurrences within the query results:
+
+1. In the query results view, press <kbd>Cmd/Ctrl</kbd>&nbsp;<kbd>F</kbd>.
+1. Type the expression that you want to search for.
+1. Press <kbd>Enter</kbd> to highlight the matches and jump to the first match.
+1. Press <kbd>Enter</kbd> repeatedly to go forward in the list of matches, and press <kbd>Enter</kbd>&nbsp;<kbd>Shift</kbd> to go backward.
+
+Currently, the search only highlights matching entries in the events rendered on your screen.
```
gabrielelpidio: What Dom said

query-data/explore.mdx
```diff
@@ -246,4 +246,15 @@ In the event timeline, line charts, and heat maps, you can drag the pointer over
...
\ No newline at end of file
+The time range of your query automatically updates to match what you selected.
+
+### Search within query results
+
+To quickly search for an expression and highlight its occurrences within the query results:
+
+1. In the query results view, press <kbd>Cmd/Ctrl</kbd>&nbsp;<kbd>F</kbd>.
+1. Type the expression that you want to search for.
+1. Press <kbd>Enter</kbd> to highlight the matches and jump to the first match.
```
gabrielelpidio: The first match should get jumped on automatically, enter/shift enter are only for navigation

gabrielelpidio (changes requested):

query-data/explore.mdx
gabrielelpidio: > It’s the opposite of this. Using browser search would be limited to highlighting matching entries in events rendered on screen since the list is virtualized, but Axiom’s `Cmd/Ctrl+F` search feature will return every relevant search response in the entire virtualized list (i.e. it will match entries from all results returned by the query - respecting automatic limits like 1,000).

gabrielelpidio (approved):

PR #248, opened 2025-03-12
Add externaldata

manototh:


apl/tabular-operators/externaldata-operator.mdx
```diff
@@ -0,0 +1,106 @@
+---
+title: externaldata
+description: 'This page explains how to use the externaldata operator in APL.'
+---
+
+The `externaldata` operator in APL allows you to retrieve data from external storage sources, such as Azure Blob Storage, AWS S3, or HTTP endpoints, and use it within queries. You can specify the schema of the external data and query it as if it were a native dataset. This operator is useful when you need to analyze data that is stored externally without importing it into Axiom.
```
gordallott, line 6: > This operator is useful when you need to analyze data that is stored externally without importing it into Axiom.

generally externaldata is more used to add extra information into an existing query, if people want to analyse data that isn't in axiom... they should get it in axiom and just query it the normal way. 

an example use-case might be that you have a dataset with access logs in it, where the access logs have a field 'employeeID'. 

you then might want to add extra information into your query by cross-referencing each employeeID with an employeeID defined in a csv hosted somewhere else. Lets say you have a csv like 
```csv
employeeID, email, name, location
00001, tifa@acme.com, Tifa Lockhart, america
00002, barret@acme.com, Barret Wallace, europe
00003,  cid@acme.com, Cid Highwind, europe
```

and a query like 
```kql
let employees=externaldata(employeeID: string, email: string, name: string, location: string)["http://wherever.com/the/csv/is.csv"] with (format="csv", skipFirstRow=true);
accessLogs
| where severity == "high"
| lookup employees on employeeID
| project resource, severity, employeeID, email, name // the last two come from the csv
```

apl/tabular-operators/externaldata-operator.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+<Note>
+The `externaldata` operator currently supports external data sources with a file size of maximum 5 MB.
+
+In queries with `externaldata`, reference at least one Axiom dataset. Axiom doesn’t support APL queries that don’t reference Axiom datasets.
```
gordallott: this might change before release, will update if it does

gordallott (approved):

apl/tabular-operators/externaldata-operator.mdx
manototh: Thanks a lot! I've added this as another example. Could you please take another look? https://github.com/axiomhq/docs/pull/248/commits/55c68d016c4925be36b5dae585b8979ed46a7386

apl/tabular-operators/externaldata-operator.mdx
gordallott: it did change! 
datasetless queries work now so this can be removed, 
example

PR #250, opened 2025-03-13
Add Axiom tour

manototh:
This PR integrates the [interactive demo of Axiom](https://app.navattic.com/axiom/product-demos/cm5x5fmbx000803kycyv58wky) to the docs as a new entry in the left nav that opens in a new browser tab.

Preview: https://axiom-mano-axiom-tour.mintlify.app/introduction

## Background

We have three ways of integrating the interactive demo into the docs:
- New page in getting started: The drawback is that you can only navigate back by clicking **Documentation** which is not intuitive.
- New tab: This adds a new tab which might be too prominent for this demo.
- A new entry in the left nav that opens in a new browser tab (preferred).

I've tried the other ways of embedding the demo to the docs (like pop-up), but they didn't work for me.


dominicchapman (approved):
Looks great - let's get this one merged!

PR #268, opened 2025-04-24
Update information on regions

manototh:
Previews:
https://axiom-mano-regions-update.mintlify.app/reference/regions
https://axiom-mano-regions-update.mintlify.app/send-data/ingest#regions
https://axiom-mano-regions-update.mintlify.app/send-data/fluent-bit

### DIFF
--- a/reference/regions.mdx
+++ b/reference/regions.mdx
@@ -11,18 +11,20 @@ The examples in this documentation use the US domain. If your organization uses 
 
 ## Check your region
 
-To check which region your organization uses, open the Axiom web app and check the URL in the browser:
-- If the URL starts with `https://app.axiom.co/`, you use the default US region.
-- If the URL starts with `https://app.eu.axiom.co/`, you use the EU region.
+To check which region your organization uses, match the base domain in the Axiom web app with the table below:
 
-## Axiom app
-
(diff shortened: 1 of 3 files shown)


benenharrington:
I'm still struggling with the thought that these changes would have made no difference to the user that got stuck with the fluentbit code sample. In the code snippet we set the value of `Host` to `api.axiom.co`, which doesn't match the pattern you're documenting here. 

You mentioned the possibility of also making a change to the Fluent Bit page that made it clear that the value of Host would depend on the user's region. Can we include that in this PR? I promised the user who raised this issue that I'd follow up to let them know about changes we made to make this issue less likely in the future, and I don't really feel like I can do that with these changes.

manototh:
@benenharrington I think there were many problems with this page so I had to substantially rewrite it. Please let me know if this helps: https://axiom-mano-regions-update.mintlify.app/send-data/fluent-bit

benenharrington (approved):

benenharrington:
> @benenharrington I think there were many problems with this page so I had to substantially rewrite it. Please let me know if this helps: https://axiom-mano-regions-update.mintlify.app/send-data/fluent-bit

Thanks Mano! Looking forward to getting back to trackunit with the changes

PR #272, opened 2025-04-28
Provide better guidance on regions and domains

manototh:
It's a huge PR but most of the changes are trivial. The only places that require in-depth review are where I change the env vars:
https://github.com/axiomhq/docs/pull/272/files#diff-dd92450479286bfa504190cafbff89c9499435e875578aa551f8b76da8f0a904
https://github.com/axiomhq/docs/pull/272/files#diff-f1ede597e9e2d9be2dee268ba140c4a21a6173fe18fb8af5ea35013ceda0c811
https://github.com/axiomhq/docs/pull/272/files#diff-21b88616797458994322b055f38a94ec640bd9780e241fed25c46efa54648dea
https://github.com/axiomhq/docs/pull/272/files#diff-2a80cb23e3b06ba82d128786b27121e9e2062d6a02438c44ed4cfc0ab3cd5008


dominicchapman (commented):
Nice!
I wonder if `AXIOM_DOMAIN` would be more consistent with other placeholders?
Are there other spots we will cover too? For example, the endpoint URL here: https://axiom.co/docs/send-data/opentelemetry

dominicchapman (approved):
We might consider using infobox if we want to bring a bit more visual prominence to this. Otherwise, looks great. A solid way to set us up for e***s! :-)
image

thecraftman (approved):

manototh:
@dominicchapman Thanks, good idea! What do you think about an info box with white background? It makes the text stand out but it doesn't make it overly prominent.

image

https://axiom-mano-unify-domains.mintlify.app/guides/opentelemetry-django#create-exporter-py-for-tracer-configuration
https://axiom-mano-unify-domains.mintlify.app/guides/opentelemetry-cloudflare-workers#cloudflare-workers-script-configuration-index-ts

dominicchapman:
> @dominicchapman Thanks, good idea! What do you think about an info box with white background? 

Looks good to me! :-)

PR #273, opened 2025-04-29
Add column_ifexists

manototh:
Preview: https://axiom-mano-column-ifexists.mintlify.app/apl/scalar-functions/metadata-functions/column-ifexists


apl/scalar-functions/conditional-functions/column-ifexists.mdx
```diff
@@ -0,0 +1,150 @@
...
+| project _time, trace_id, span_id, status
+```
+
+[Run in Playground](https://play.axiom.co/axiom-play-qf1k/query?initForm=%7B%22apl%22%3A%22%5B'otel-demo-traces'%5D%20%7C%20extend%20status%20%3D%20column_ifexists('status_code'%2C%20'OK')%20%7C%20project%20_time%2C%20trace_id%2C%20span_id%2C%20status%22%7D)
+
+**Output**
+
+| _time                | trace_id     | span_id     | status |
+|---------------------|--------------|-------------|--------|
+| 2025-04-28T10:30:12Z | abc123       | span567     | nil  |
```
ted-gould: This `nil` should be `"OK"`. And probably, we shouldn't mix types here. That always gets confusing. Probably better to return 100 or something.

manototh: @ted-gould Thanks for this. I have changed the fallback to `100` as you suggested. But this is the fallback if the field doesn't exist. In the example, some values are missing from the field and these are returned as `nil` so I think the table is correct. Could you please confirm?

ted-gould: Ah, okay. Sorry I was confusing with `coalesce`. But we should probably have something in the example where we're using the feature of the function. Perhaps something like `extend otherstatus = column_ifexists('status_code_notexist', 100)` so that we show it working.

manototh: @ted-gould Many thanks. I've added this example here: https://axiom-mano-column-ifexists.mintlify.app/apl/scalar-functions/metadata-functions/column-ifexists#open-telemetry-traces

ted-gould (approved):

PR #264, opened 2025-04-10
Add new Datasets tab details, create map fields

manototh:

PR #275, opened 2025-04-30
Add views

manototh:
Preview: https://axiom-mano-views.mintlify.app/query-data/views

Prerequisite: https://github.com/axiomhq/docs/pull/264


query-data/datasets.mdx
```diff
@@ -55,6 +56,38 @@ For more information, see [Virtual fields](/query-data/virtual-fields).
 
 Map fields are a special type of field that can hold a collection of nested key-value pairs within a single field. You can think of the content of a map field as a JSON object. The Dataset tab enables you to create map fields, and view unused and removed map fields. For more information, see [Map fields](/apl/data-types/map-fields#create-map-fields).
 
+## Views
+
+Views allow you to grant other users scoped access to your datasets. This means that instead of sharing the whole dataset with other users, you have the option to share only a filtered and transformed representation of the data. You define a view using a query that applies filters and transformations to your dataset, and then you only share the results of this query with another user.
```
benenharrington: I think this describes one use case for Views, but [there are certainly others](https://www.notion.so/axiomhq/Views-06f0d9bfc0ab4f89bd3393b6db265088). The goal of the suggestion below is to describe the functionality at a high level, in a way that's relevant for all the use cases we have in mind. Very open to other ways of approaching this.

```suggestion
Views allow you to apply commonly-used filters and transformations to your dataset, and to treat the resulting output like you would any other dataset. The concept of a View in Axiom is similar to the concept of a virtual table in a database.
```

query-data/datasets.mdx
```diff
@@ -55,6 +56,38 @@ For more information, see [Virtual fields](/query-data/virtual-fields).
...
+1. In the top right of the **Views** section, click **+ Create view**.
+1. Create the query using the [visual query builder](/query-data/explore#create-a-query-using-the-visual-query-builder) or the [Axiom Processing Language (APL)](/query-data/explore#create-a-query-using-apl).
+1. Below the query, check the preview of the events and the fields that your query returns.
+1. Name the view and optionally add a description.
+1. Click **Create view**.
+
+### Display view
+
+1. Select a dataset.
+1. In the **Views** section on the right, click the view you want to display.
```
benenharrington: I'm not sure this section is of much value

query-data/datasets.mdx
```diff
@@ -55,6 +56,38 @@ For more information, see [Virtual fields](/query-data/virtual-fields).
...
+1. Click <img src="/doc-assets/icons/settings.svg" className="inline-icon" alt="Settings icon" /> **Settings > Roles**.
+1. Click the role with which you want to share the view or create a new role.
+1. In the **Organization** section, enable read access to views.
+1. In the **Individual datasets** section, click **Add dataset**. 
+1. Start typing the name of the view. In the list, find the view that you want to share, and then click **Add**.
+    <Note>
+        It’s enough to grant access to the view. You don’t need to grant access to the underlying dataset.
+    </Note>
+1. Click **Save**.
+
```
benenharrington: Similar the comment above, I think this is focussing too much on a particular use case of views.

manototh:
@benenharrington Many thanks for this. I see your point and I've addressed your suggestions: https://axiom-mano-views.mintlify.app/query-data/views

I think it's worth keeping the procedures, especially the one about sharing views, because it explains something that isn't intuitive in our UI at the moment.

query-data/views.mdx
```diff
@@ -0,0 +1,45 @@
...
+1. Click **Create view**.
+
+### Display view
+
+1. Select a dataset.
+1. In the **Views** section on the right, click the view you want to display.
+
+### Grant access to view
+
+Grant other users access to a view in the same way you grant access to individual datasets:
```
benenharrington:
```suggestion
All views are available to users with organization-level read access to views. You can allow users without these permissions to access an individual view in the same way you grant access to individual datasets:
```

benenharrington:
Looks great Mano!

Added one small suggestion to make it explicit that certain users will have access to a view as soon as it's created (similar to datasets). Feel free to ignore if you don't think it's necessary

query-data/views.mdx
manototh: Thanks @benenharrington!
Please feel free to commit this if my rewrite works for you.

query-data/views.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+1. Click <img src="/doc-assets/icons/settings.svg" className="inline-icon" alt="Settings icon" /> **Settings > Roles**.
+1. Click the role with which you want to share the view or create a new role.
+1. In the **Organization** section, enable read access to views.
```
manototh: @benenharrington This should be removed, right?

benenharrington: Woops, yep!

benenharrington (approved):

PR #277, opened 2025-05-06
docs: Move Join operator to public preview

dominicchapman:
- Moves `join` operator to public preview rather than private preview
- Fixes references to "Features states" instead of "Feature states" across docs
- Revises limits for `join` operator to reflect more recent update (graduated limits, based on field count)

Primary pages to preview are:
1. https://axiom-dominic-join-feature-state.mintlify.app/apl/tabular-operators/join-operator
2. https://axiom-dominic-join-feature-state.mintlify.app/getting-started-guide/feature-states


kamstrup (commented):
Looks good :+1: 

Somewhat related to this PR: I think we only support `leftouter`, `inner`, and `innerunique` join types. The docs list all the ones that kusto support.

Adding a few more join types would be 1-2 weeks of work I think. In case we want to commit to it.

getting-started-guide/feature-states.mdx
```diff
@@ -22,13 +22,13 @@ Private and public preview features are experimental, are not guaranteed to work
...
 
 - [Flow](/process-data/introduction)
-- [Join operator](/apl/tabular-operators/join-operator)
 - [Progressive query mode](/query-data/explore#query-modes)
 - [Send data from Datadog to Axiom](/send-data/datadog)
 
 Current public preview features:
 - [Cursor-based pagination](/restapi/pagination)
 - [externaldata operator](/apl/tabular-operators/externaldata-operator)
+- [Join operator](/apl/tabular-operators/join-operator)
```
manototh:
```suggestion
- [join operator](/apl/tabular-operators/join-operator)
```

apl/tabular-operators/join-operator.mdx
```diff
@@ -12,9 +12,9 @@ This operator is useful when you want to:
 - Enrich existing data with supplementary details.
 
 <Note>
-    The `join` operator is currently in private preview. To try it out, [contact Axiom](https://axiom.co/contact).
+    The `join` operator is currently in public preview. For more information, see [Feature states](/getting-started-guide/feature-states).
 
-    The preview of the `join` operator works with an upper limit of 5,000 events on the left side of the join and 50,000 on the right side of the join.
+    The preview of the `join` operator works with variable limits depending on the structure of your dataset. For the left side of the join, the limit is 50,000 rows when the dataset has fewer than 100 fields. This limit decreases linearly as the field count increases: 25,000 rows at 200 fields, 12,500 rows at 400 fields, and 10,000 rows at 500+ fields. The right side of the join has a consistent limit of 50,000 rows.
```
manototh:
```suggestion
    The preview of the `join` operator works with variable limits depending on the structure of your dataset. For the left side of the join, the limit decreases linearly as the field count increases:
        - 50,000 rows for datasets with fewer than 100 fields
        - 25,000 rows for datasets with up to 200 fields
        - 12,500 rows for datasets with up to 400 fields
        - 10,000 rows for datasets with 500 fields or more
    The right side of the join has a consistent limit of 50,000 rows.
```

manototh: Thanks for finding out these details. I have converted it to a list for better readability but a table might work even better.

I was wondering about the cut-off points. The original text said `the limit is 50,000 rows when the dataset has fewer than 100 fields ... 25,000 rows at 200 fields, 12,500 rows at 400 fields, and 10,000 rows at 500+ fields.` Am I right to interpret this that the limit changes to 25,000 rows at 200 fields? If yes, the limit is 50,000 rows for datasets with up to 200 rows, right?
My other interpretation is that the limit is 25,000 rows for datasets with 101 to 200 fields. But then I'm not sure what happens between 400 and 500. Should I find this out or do you already know the answer?

manototh (approved):

manototh:
> Looks good 👍
> 
> Somewhat related to this PR: I think we only support `leftouter`, `inner`, and `innerunique` join types. The docs list all the ones that kusto support.
> 
> Adding a few more join types would be 1-2 weeks of work I think. In case we want to commit to it.

@kamstrup Thanks for this. My idea was to give a general overview of join as a concept. That's why I explain the different kinds of join. The docs currently says that `The preview of the join operator currently only supports inner join. Support for other kinds of join is coming soon.` If it's no longer true and we don't have such plans anymore, I'll update it.

manototh:
@dominicchapman @kamstrup Thanks a lot for this PR! I've just had a question about interpreting the cut-off points, please. But this should not block the publication in case this is urgent.

apl/tabular-operators/join-operator.mdx
kamstrup: It is a smooth linear interpolation, so it doesn't jump in steps. Ie, if you pick some random number like 287 fields, then it'll be adjusted to that.

A table with some other numbers:
```
numFields vs dynamicMaxLimit
1 50000
34 50000
67 50000
100 50000
133 37593
166 30120
199 25125
232 21551
265 18867
298 16778
331 15105
364 13736
397 12594
430 11627
463 10799
496 10080
```

manototh: Thanks a lot! I don't know if users expect this level of specificity, but I'm happy to put this in the docs if this is the case. What do you think @dominicchapman ?

kamstrup: you text also says that it decreases linearly, so I think it is already correct

manototh: Ah, I get it now. The original text only gave some examples. In this case, it's fine, I would just add a short thing: https://github.com/axiomhq/docs/pull/277/commits/48eb716a22d26a836269ac3e8b4ef45b90394df3

PR #278, opened 2025-05-12
fix-aws-s3-lambda-code

thecraftman:
- update code to support different formats
- check for `.log` format where you have elb logs


send-data/aws-s3.mdx
```diff
@@ -62,9 +63,37 @@ def lambda_handler(event, context):
...
+                    log_lines = log_data.strip().split("\n")
+                    json_logs = []
+                    for line in log_lines:
+                        try:
+                            # Try to parse each line as JSON
+                            parsed_line = json.loads(line)
+                            json_logs.append(parsed_line)
+                        except:
+                            # If not JSON, treat as plain text
+                            json_logs.append({'message': line})
```
darach: The line, even if plain text, needs to escape any internal quotes or other characters as it is being converted to a utf-8 quoted string so that it can be forwarded as a json log line. This would result in nginx or syslog data being malformed due to the implicit conversion not handling data that should be escaped...

darach (changes requested):

send-data/aws-s3.mdx
thecraftman: @darach fixed as well.

thesollyz (approved):

PR #298, opened 2025-05-28
Add note for allowed APL operators for Flow

benenharrington:
The complete list is 
```
where
search
project
project-away
project-keep
project-rename
project-reorder
extend
extend-valid
redact
parse
parse-kv
```
It feels simpler to refer to these as 'filter and transformation' operators but open to alternatives.


manototh (approved):

manototh:
@benenharrington Thanks for this!

PR #65, opened 2024-09-12
Add docs about map fields

c-ehrlich:
Preview: https://axiom-map-fields-docs.mintlify.app/apl/data-types/map-fields

PR #304, opened 2025-06-03
Update map fields

manototh:
Preview: https://axiom-mano-fix-map-fields.mintlify.app/apl/data-types/map-fields


apl/data-types/map-fields.mdx
```diff
@@ -73,4 +71,29 @@ To view map fields:
...
\ No newline at end of file
+- **REMOVED** means that the field was configured as a map field but at some point it was changed to an ordinary field. Axiom adds new events to the field as usual. Events you ingested before the change retain the map structure. To fully remove this field, first [trim your dataset](/reference/datasets#trim-dataset) to remove the time period when map data was ingested, and then [vacuum the fields](/reference/datasets#vacuum-fields).
+
+## Map fields and flattened fields
+
+The parent map field and its flattened subfields can appear in the same dataset. For example, ingesting the following JSON adds a map field (`geo`) and two flattened fields (`geo.city`, `geo.country`) to the dataset. Axiom treats them as separate fields and doesn’t maintain a relationship between them.
```
flbn: ingesting this doesn't create a map field; just the two flattened fields. to create a map field, you can do so via the api or now, in app.

the point about axiom treating them differently is if you have a map field that shares the same keys/values as a flattened version. theres no conflicts there, like you mentioned

apl/data-types/map-fields.mdx
```diff
@@ -73,4 +71,29 @@ To view map fields:
...
+    "city": "Paris",
+    "country": "France"
+  }
+}
+```
+
+This means the following:
+
+- Projecting away all subfields (in this example, `geo.city` and `geo.country`) doesn’t remove the parent map field `geo` unless it is explicitly projected away.
+- Fields metadata may still include entries for projected-away fields if the map field is preserved.
```
flbn: i wouldn't note this, those are implementation details i needed at the time but customers dont see it

apl/data-types/map-fields.mdx
```diff
@@ -73,4 +71,29 @@ To view map fields:
...
+- Queries using `['geo.city']` access a field literally named `geo.city`, while `['geo']['city']` accesses the `city` key inside a `geo` map. These references are not equivalent.
```
flbn: ❤️

BOLD this lolol

apl/data-types/map-fields.mdx
```diff
@@ -6,11 +6,7 @@ tags: ['axiom documentation', 'documentation', 'axiom', 'map fields', 'object fi
 
 Map fields are a special type of field that can hold a collection of nested key-value pairs within a single field. You can think of the content of a map field as a JSON object.
 
-<Note>
-Currently, Axiom automatically creates map fields in datasets that use [OpenTelemetry](/send-data/opentelemetry). You cannot create map fields yourself.
-
-Support for creating your own map fields is coming in early 2025. To express interest in the feature, [contact Axiom](https://axiom.co/contact).
-</Note>
+Axiom automatically creates map fields in datasets that use [OpenTelemetry](/send-data/opentelemetry) and you can create them yourself.
```
c-ehrlich:
```suggestion
Axiom automatically creates the `attributes.custom` map field in datasets that use [OpenTelemetry](/send-data/opentelemetry). You can also create other map fields in any dataset.
```

the current "and you can create them yourself" feels unclear imo. not sure my suggestion is exactly right though.

apl/data-types/map-fields.mdx
c-ehrlich: yes ollie is right about this

the steps to end up in this situation are:

1. `geo` is not a map field
2. ingest

```json
{
  "geo": {
    "city": "Paris",
    "country": "France"
  }
}
```

this gets put into `geo.city` and `geo.country`

3. make `geo` a map field
4. ingest the same


```json
{
  "geo": {
    "city": "Paris",
    "country": "France"
  }
}
```

this gets put into `geo`

(or in opposite order)

apl/data-types/map-fields.mdx
c-ehrlich: i agree, the information below about `['geo']['city'] being different from ['geo.city'] is enough here. imo the above only makes it more confusing to someone who is reading about this for the first time.

c-ehrlich (approved):

apl/data-types/map-fields.mdx
manototh: Thanks! I think it's not just `attributes.custom` that we create in OTel datasets. I would not specify the actual fields here. But the second part of the sentence could surely be improved, so implemented.

c-ehrlich: afaik that's the only one. which other ones do you know of?

apl/data-types/map-fields.mdx
```diff
@@ -69,4 +56,64 @@ To view map fields:
...
+To access the properties of nested maps, use dot notation, index notation, or a mix of the two. If you use index notation for an entity, enclose the entity name in quotation marks (`'` or `"`) and square brackets (`[]`). For example:
+
+- `where map_field.property1.property2 == 14`
+- `where ['map_field'].property1.property2 == 14`
+- `where ['map_field']['property1']['property2'] == 14`
+
+If an entity name has spaces (` `), dots (`.`), or dashes (`-`), you can only use index notation for that entity. You can use dot notation for the other entities. For example:
+
+- `where ['map.field']['property.name1']['property.name2'] == 14`
+- `where ['map.field'].property1.property2 == 14`
```
c-ehrlich: looks like this was already merged in a previous pr. but as far as i can tell all of these apl examples will not return the desired events.

the only correct syntax is `['map.field']['something.inside.of.it']`

see: /query?qid=4O0s3Ki3fPa-sxyen5

aside: imo even if there is some other way that works, we should only teach `['map.field']['something.inside.of.it']`.

c-ehrlich (changes requested):

apl/data-types/map-fields.mdx
manototh: I thought `resource.custom` could also be one of them? Anyway, it's OK not to specify it here. We do it elsewhere.

apl/data-types/map-fields.mdx
manototh: @c-ehrlich Thanks for this. I don't know if this recently got broken because the original PR underwent a thorough review process: https://github.com/axiomhq/docs/pull/65

I've tried to implement your suggestions here: https://github.com/axiomhq/docs/pull/304/commits/61e91816afec32013fb910b77cb5718c023ee8ca

The problem is that `where ['map.field']['property.name1']['property.name2'] == 14` still doesn't work but then I don't know how else to deal with property names with special characters like dots.

manototh: @c-ehrlich I have updated this section based on [Michal's comment](https://watchlyhq.slack.com/archives/C083K4B8DUY/p1750408130978639?thread_ts=1750407799.992579&cid=C083K4B8DUY): https://axiom-mano-fix-map-fields.mintlify.app/apl/data-types/map-fields#access-properties-of-nested-maps

Please let me know if this is OK. All the variations work on /query?qid=pDnbJ4Db6nO-sy5dyh&relative=1

c-ehrlich (approved):

PR #320, opened 2025-06-30
Use keywords meta tag to improve internal search

manototh:
Use keywords meta tag based on https://mintlify.com/docs/settings/seo


manototh:
Closing due to flawed approach https://mintlify.com/docs/pages#internal-search-optimization

PR #332, opened 2025-07-08
Add docs satisfaction survey to each page

manototh:
https://axiom-mano-survey-js.mintlify.app/introduction
Appears after 10 seconds


dominicchapman (approved):
Excited for this! Nice one :-)

PR #340, opened 2025-07-13
Update Cloudflare Workers Domain string with backticks

ahmadaccino:
Since you're doing string interpolation this string needs back tips like the bearer token string does


manototh (approved):

manototh:
@ahmadaccino Many thanks for this. Well spotted!

PR #346, opened 2025-07-18
fix: revise docs based on user experience

dominicchapman:
- Was kind of annoying not having a quick way to install the dependencies listed in the quickstart
- I was getting warnings about use of `Resource` (a type) - looks like OTel docs use `resourceFromAttributes` and this works in my testing
- The `tracer` install was incorrect. Should be being created by user. Examples do it in separate file but seems redundant for this

### DIFF
--- a/ai-engineering/quickstart.mdx
+++ b/ai-engineering/quickstart.mdx
@@ -50,20 +50,70 @@ The Axiom AI SDK is built on the OpenTelemetry standard and requires a configure
 
 Here is a standard configuration for a Node.js environment:
 
+<CodeGroup>
+
+```bash pnpm
+pnpm i \
+  dotenv \
+  @opentelemetry/exporter-trace-otlp-http \
+  @opentelemetry/resources \
+  @opentelemetry/sdk-node \
+  @opentelemetry/sdk-trace-node \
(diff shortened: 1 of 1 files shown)


ai-engineering/quickstart.mdx
```diff
@@ -50,20 +50,70 @@ The Axiom AI SDK is built on the OpenTelemetry standard and requires a configure
 
 Here is a standard configuration for a Node.js environment:
 
+<CodeGroup>
+
+```bash pnpm
+pnpm i \
+  dotenv \
```
c-ehrlich, line 57: also `ai` and `@axiomhq/ai`?

c-ehrlich (approved):
All of these changes look correct (except also needing `ai`/`@axiomhq/ai`). Apologies for not catching the first time.

ai-engineering/quickstart.mdx
dominicchapman: Thanks! `@axiomhq/ai` is included higher up on the Quickstart page so those following will have installed. `ai` package is only needed in other sections like Observe where it might be worth doing something like this changeset. Will this update for now!

PR #350, opened 2025-07-24
Introduce tool calling to AI engineering

dominicchapman:
- Fixes instrumentation code block used in two pages - made consistent with docs snippet
- Introduces `wrapTool`
- Adds code group to show other model providers (OpenAI, Anthropic, Google Gemini, Grok)

### DIFF
--- a/ai-engineering/observe.mdx
+++ b/ai-engineering/observe.mdx
@@ -23,7 +23,39 @@ The easiest way to get started is by wrapping your existing AI model client. The
 
 The `wrapAISDKModel` function takes an existing AI model object and returns an instrumented version that will automatically generate trace data for every call.
 
-```typescript
+<CodeGroup>
+
+```typescript OpenAI
+// src/shared/openai.ts
+
+import { createOpenAI } from '@ai-sdk/openai';
+import { wrapAISDKModel } from '@axiomhq/ai';
+
+const openaiProvider = createOpenAI({
+  apiKey: process.env.OPENAI_API_KEY,
+});
+
+// Wrap the model to enable automatic tracing
+export const gpt4o = wrapAISDKModel(openaiProvider('gpt-4o'));
+export const gpt4oMini = wrapAISDKModel(openaiProvider('gpt-4o-mini'));
+```
+
+```typescript Anthropic
+// src/shared/anthropic.ts
+
+import { createAnthropic } from '@ai-sdk/anthropic';
+import { wrapAISDKModel } from '@axiomhq/ai';
+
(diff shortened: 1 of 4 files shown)


ai-engineering/observe.mdx
```diff
@@ -22,7 +23,39 @@ The easiest way to get started is by wrapping your existing AI model client. The
...
+
+```typescript OpenAI
+// src/shared/openai.ts
+
+import { createOpenAI } from '@ai-sdk/openai';
+import { wrapAISDKModel } from '@axiomhq/ai';
+
+const openaiProvider = createOpenAI({
+  apiKey: process.env.OPENAI_API_KEY,
+});
```
c-ehrlich, line 36: when using v4 of the vercel sdk, `compatibility: strict` is required for streaming to be instrumented properly. see: https://github.com/axiomhq/ai/blob/main/examples/example-instrumentation-nextjs-v4/src/shared/openai.ts

i believe this is only true for the openai provider, but obviously haven't tested ALL of them

ai-engineering/observe.mdx
```diff
@@ -71,46 +121,118 @@ export default async function Page() {
...
+// In your generateText call, provide wrapped tools
+const { text, toolResults } = await generateText({
+  model: gpt4o,
+  messages: [
+    { role: 'system', content: 'You are a helpful assistant.' },
+    { role: 'user', content: 'How do I get from Paris to Berlin?' },
+  ],
+  tools: {
+    // Wrap each tool with its name
+    findDirections: wrapTool(
```
c-ehrlich, line 147: note: if you're using multiple tools there's also `wrapTools` which works like 

```ts
await generateText({
  model,
  messages,
  tools: wrapTools(
    findDirections: tool({ /* args */ }),
    someOtherTool: tool({ /* args */ }),
  ),
})
```

maybe we even want to present that as the MAIN way of wrapping tools? the nice thing is it doesn't require passing the name as a string because it can figure it out from the shape of the object

(the plus side of `wrapTool` is that if you define your tools in some other file, you can wrap them before export, or maybe you don't want to wrap all tools because some of them do custom stuff)

c-ehrlich (approved):

ai-engineering/observe.mdx
dominicchapman: Is this true for `v5` also?

ai-engineering/observe.mdx
dominicchapman: Excellent guidance, will incorporate in follow up PR - resolving to get this portion merged

PR #354, opened 2025-07-30
Update info about compression of map fields

manototh:


apl/data-types/map-fields.mdx
```diff
@@ -22,7 +22,6 @@ Use map fields in the following cases:
 
 Map fields reduce impact on field limits, but involve trade-offs in query efficiency and compression. The drawbacks of map fields are the following:
 - Querying map fields uses more query-hours than querying conventional fields.
-- Map fields don’t compress as well as conventional fields. This means datasets with map fields use more storage.
```
tsenart, line 25: It's a bit more nuanced than that. Map fields will compress nearly as well as regular fields when there's a lot of shared structure / fields between map values. If every map value is completely different, different keys, etc, then for sure they won't compress so well. When the maps are mostly the same structure, they'll compress better.

Now, apart from compression, there's a lot more data in a map than, say, and integer, or a small string. So, while we will compress well the common sub fields across maps, the top level map field overall will still consume storage proportionally to the number of fields it has and the shape of its values.

apl/data-types/map-fields.mdx
```diff
@@ -22,7 +22,6 @@ Use map fields in the following cases:
 
 Map fields reduce impact on field limits, but involve trade-offs in query efficiency and compression. The drawbacks of map fields are the following:
 - Querying map fields uses more query-hours than querying conventional fields.
```
manototh, line 24:
```suggestion
- Querying map fields uses more query-hours than querying conventional fields.
- In some cases, map fields don’t compress as well as conventional fields. For example, if there is little shared structure between map values. This means datasets with map fields can use more storage.
```

apl/data-types/map-fields.mdx
manototh: @tsenart Many thanks for this. I've tried to capture the essence of these nuances. Please let me know if this works better.

tsenart (approved):

PR #356, opened 2025-08-04
Add named fields to argmin and argmax

manototh:
Check links to Playground and example output after release


gordallott (approved):
:100:

gordallott:
just noting that the deploy last week didn't go out, it'll go out this week (starting 11th august)

PR #364, opened 2025-08-13
Add Axiom regions to AWS Lambda

ipapast:
There is no mention for the Region setting in all docs, so it's easy to miss this. 

In our case, we were getting 403: Insufficient permissions as a result.


thesollyz (approved):

manototh:
@ipapast Many thanks for your valuable contribution!

PR #370, opened 2025-08-20
Add reserved field names, clarify action if hitting field limits

manototh:
Preview: https://axiom-mano-reserved-fields.mintlify.app/reference/field-restrictions#limits-on-ingested-data

Basis:
https://watchlyhq.slack.com/archives/C03832MLBJA/p1755690031802139
https://watchlyhq.slack.com/archives/C1NJU9Y6B/p1756116638577429?thread_ts=1755726109.359199&cid=C1NJU9Y6B


reference/datasets.mdx
```diff
@@ -69,7 +69,20 @@ Axiom creates the following two fields automatically for a new dataset:
...
+In most cases, use `_time` to define the timestamp of events. In rare cases, if you experience clock skews on your event-producing systems, `_sysTime` can be useful.
+
+### Reserved field names
+
+Axiom reserves the following field names for internal use:
+
+- `_cursor`
+- `_rowID`
+- `_source`
+- `_sysTime`
```
kamstrup: please add `_blockInfo` here as well

kamstrup (approved):
Thanks! <3

reference/field-restrictions.mdx
```diff
@@ -45,15 +43,57 @@ For more information on how to save on data loading, data retention, and queryin
...
+
+|                              | Limit     |
+| ---------------------------- | --------- |
+| Maximum event size           | 1 MB      |
+| Maximum events in a batch    | 10,000    |
+| Maximum field name length    | 200 bytes |
+
+If you try to ingest data that exceeds these limits, Axiom does the following:
+- Replaces strings that are too long with `<invalid string: too long>`.
+- Replaces binary data that’s too long with `<invalid data>`.
```
kamstrup:
```suggestion
- Replaces binary with `<invalid data>`.
```

reference/field-restrictions.mdx
```diff
@@ -45,15 +43,57 @@ For more information on how to save on data loading, data retention, and queryin
...
 
 ### Limits on ingested data
 
-<IngestDataLimits />
+The table below summarizes the limits Axiom applies to each data ingest. These limits are independent of your pricing plan.
+
+|                              | Limit     |
+| ---------------------------- | --------- |
+| Maximum event size           | 1 MB      |
+| Maximum events in a batch    | 10,000    |
```
kamstrup, line 63: I don't think we have a limit on the number of events in a batch?

Maybe also list maximum field size of 1 MB? It feels a bit tautological (because the event itself cannot be bigger), but it makes sense to mention in the case where we replace strings longer than this with `<invalid string: too long>`

reference/field-restrictions.mdx
```diff
@@ -45,15 +43,57 @@ For more information on how to save on data loading, data retention, and queryin
...
+- Converts the following float values to `nil`:
+    - NaN
+    - +Infty
+    - -Infty
+
+### Special fields
+
+Axiom creates the following two fields automatically for a new dataset:
+
+- `_time` is the timestamp of the event. If the data you ingest doesn’t have a `_time` field, Axiom assigns the time of the data ingest to the events.
```
kamstrup: Does it makes sense to add a link here, to how users can control what field from their data goes into `_time`?

kamstrup (approved):
All good, some small suggestions/questions

reference/field-restrictions.mdx
manototh: Thanks, good idea. How does this look? https://github.com/axiomhq/docs/pull/370/commits/216f64e3de9760c4f5b37c38e30ef87b2d4e69c2

kamstrup: yep, looks good :+1:

PR #421, opened 2025-09-26
fix(docs): cli legacy page missing redirection

antlio:
**description**
when clicking through the website mega menu cli install [link](https://axiom.co/docs/cli), we are facing a 404 page.

**solution**
this pr adds a redirection from the cli legacy page to the new one, as it seems to be used within the website mega menu:

| preview |
|------|
| image |
| the link being used within the menu is pointing to `docs/cli` which is throwing 404 | 

another take could be to update the link within the website also, but redirection might do the trick to ensure availability if used elsewhere.

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -1042,6 +1042,10 @@
     {
       "source": "/process-data/*",
       "destination": "/introduction"
+    },
+    {
+      "source": "/cli",
+      "destination": "/reference/cli"
     }
   ]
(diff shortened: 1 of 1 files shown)


manototh:
@antlio  Many thanks for this contribution. We appreciate it a lot. 🎉

manototh (approved):

PR #429, opened 2025-10-03
Fix: Update Axiom CLI Homebrew Instructions

keithf123r:
Updated Homebrew installation instructions for Axiom CLI to use --cask option which is the correct option


manototh:
@keithf123r Many thanks for this contribution. We appreciate it a lot. 🎉

This looks good to me. I've also assigned one of our Engineers for review.

ronoc (approved):

PR #462, opened 2025-11-06
Rewrite string functions

manototh:


gordallott (approved):
there's too much here to be sure, I didn't see anything that might be wrong, but i can't check everything exactly against the implementation like I would do normally fingers crossed!

PR #465, opened 2025-11-11
Update Measure in AI eng

c-ehrlich:


ai-engineering/measure.mdx
```diff
@@ -1,34 +1,45 @@
...
+  'Levenshtein',
+  ({ output, expected }: { output: string; expected: string }) => {
+    // Calculate Levenshtein distance score
+    const distance = calculateLevenshtein(output, expected);
+    const maxLen = Math.max(output.length, expected.length);
+    return maxLen === 0 ? 1 : 1 - distance / maxLen;
+  }
+);
 
 Eval('text-match-eval', {
```
islameldigwi, line 44: Should we encourage users to add capability name to metadata?

ai-engineering/measure.mdx
```diff
@@ -46,40 +57,415 @@ Eval('text-match-eval', {
...
+```
+
+## Writing a Real-World Eval
+
+Let's build a practical evaluation for a support ticket classification system.
+
+Create an eval file `src/evals/ticket-classification.eval.ts`:
+
+```typescript
+import { experimental_Eval as Eval, Scorer } from 'axiom/ai/evals';
```
islameldigwi:
```suggestion
import { Eval, Scorer } from 'axiom/ai/evals';
```

ai-engineering/measure.mdx
islameldigwi, line 1: @c-ehrlich we need to write something about cli auth: `axiom login`, maybe @manototh can help us structure this in a good way. There would be 2 ways to authenticate the SDK:
- Either using CLI, and then you only need to specify a dataset in `axiom.config.ts`
- Specifying an API token and dataset in axiom.config.ts

manototh: @thesollyz Thanks for this. I would specify API token only because that's what we use in the rest of the AI eng docs anyway.

islameldigwi: @manototh not sure what do you mean? CLI auth is a new feature that we want the users to use.

ai-engineering/measure.mdx
```diff
@@ -1,34 +1,45 @@
 ---
 title: "Measure"
 description: "Learn how to measure the quality of your AI capabilities by running evaluations against ground truth data."
-keywords: ["ai engineering", "AI engineering", "measure", "evals", "evaluation", "scoring", "graders"]
+keywords: ["ai engineering", "AI engineering", "measure", "evals", "evaluation", "scoring", "scorers"]
```
islameldigwi: I think it would be also good to have graders as a keyword here for users who would search for graders instead of scorers. Maybe we can add "scores" as well.

manototh: Good point, added

ai-engineering/measure.mdx
manototh: Do you mean CLI auth is new feature in the context of evals?

If users need to specify the API token anyway in many other places in the setup, I think it would complicate things if we asked them to authenticate the CLI as well. Or am I missing something?

ai-engineering/measure.mdx
c-ehrlich: IMO capability should be either
- parsed automatically by traversing the trace
- put somewhere more prominent than metadata, which imo is a place for the user to stick whatever they want

ai-engineering/measure.mdx
islameldigwi: yes the CLI auth is a new feature in context of evals. They don't always have to pass a token, its either:
- pass a token
- or login with CLI (then you don't need to pass a token)

c-ehrlich:
Closing in favor of https://github.com/axiomhq/docs/pull/473, feel free to re-open if that's wrong

PR #469, opened 2025-11-17
Add evals to AI engineering

manototh:
Based on https://github.com/axiomhq/docs/pull/465


ai-engineering/measure.mdx
```diff
@@ -1,85 +1,374 @@
...
+1. Follow the [Quickstart](/ai-engineering/quickstart) to set up instrumentation for your app.
+1. Run the following command to istall the `autoevals` library. This library provides pre-built scorers for common tasks like semantic similarity, factual correctness, and text matching.
 
-* `data`: An async function that returns your `collection` of `{ input, expected }` pairs, which serve as your ground truth.
-* `task`: The function that executes your AI capability, taking an `input` and producing an `output`.
-* `scorers`: An array of `grader` functions that score the `output` against the `expected` value.
-* `threshold`: A score between 0 and 1 that determines the pass/fail condition for the evaluation.
+    ```bash
+    npm install --save-dev autoevals
+    ```
```
c-ehrlich: I think it's fine if we refer to autoevals at some point in the documentation, it's a good library! But the first example should be "vanilla" IMO.

ExactMatchScorer can be rewritten as:

```ts
const ExactMatchScorer = Scorer(
  'Exact-Match',
  ({ output, expected }: { output: { response: string }; expected: { response: string } }) => {
    const normalizedOutput = output.response.trim().toLowerCase();
    const normalizedExpected = expected.response.trim().toLowerCase();

    return normalizedOutput === normalizedExpected;
    });
  }
);
```

ai-engineering/measure.mdx
```diff
@@ -1,85 +1,374 @@
...
 
-Here is an example of a complete evaluation suite:
+### Change instrumentation
 
-```ts /evals/text-match.eval.ts
-import { Levenshtein } from 'autoevals';
-import { Eval } from 'axiom/ai/evals';
+Change the instrumentation setup in the `src/instrumentation.ts` file that you have previously created in the [Quickstart](/ai-engineering/quickstart).
```
c-ehrlich: We should just tell them to define it this way in first place, then we don't need another section on `instrumentation.ts` in this page.

ai-engineering/measure.mdx
```diff
@@ -1,85 +1,374 @@
...
+  provider = new NodeTracerProvider({
+    resource: resourceFromAttributes({
+      [ATTR_SERVICE_NAME]: 'my-app', // Replace with your service name
+    },
+    {
+      // Use the latest schema version
+      // Info: https://opentelemetry.io/docs/specs/semconv/
+      schemaUrl: 'https://opentelemetry.io/schemas/1.37.0',
+    }),
+    spanProcessor: new SimpleSpanProcessor(exporter),
```
c-ehrlich: should probably be BatchSpanProcessor

ai-engineering/measure.mdx
```diff
@@ -1,85 +1,374 @@
...
+      setupAppInstrumentation({ url, token, dataset }),
+  },
+});
+```
+
+## Write evalulation function
+
+The `Eval` function provides a simple, declarative way to define a test suite for your capability directly in your codebase.
+
+The key parameters of the `Eval` function:
```
c-ehrlich, line 28: note to self: need to better document `configFlags`

ai-engineering/measure.mdx
```diff
@@ -1,85 +1,374 @@
...
+import { flag, pickFlags } from '../lib/app-scope';
+import { z } from 'zod';
+import { ExactMatch } from 'autoevals';
+
+// Define your schemas
+const ticketCategorySchema = z.enum(['spam', 'question', 'feature_request', 'bug_report']);
+const ticketResponseSchema = z.object({
+  category: ticketCategorySchema,
+  response: z.string(),
+});
```
c-ehrlich: nit: these are only used in the `schema` param of the `generateObject` call, so i would probably inline them there. having them at the top of the file makes me think they're related to the eval.

ai-engineering/measure.mdx
```diff
@@ -1,85 +1,374 @@
...
+// Custom spam classification scorer
+const SpamClassificationScorer = Scorer(
+  'Spam-Classification',
+  ({ output, expected }: { 
+    output: { category: string }; 
+    expected: { category: string };
+  }) => {
+    return (expected.category === 'spam') === (output.category === 'spam') ? 1 : 0;
+  }
+);
```
c-ehrlich, line 98: nit: this function is hard to read, would prefer something like

```ts
const SpamClassificationScorer = Scorer(
  "Spam-Classification",
  ({ output, expected }: {
    output: { category: string };
    expected: { category: string };
  }) => {
    const isSpam = (x: { category: string }) => x.category === "spam";
    return isSpam(output) === isSpam(expected) ? 1 : 0;
  }
);
```

c-ehrlich: it also feels like this is a hint that scorers should allow returning true/false, which we then parse to 1/0

ai-engineering/measure.mdx
```diff
@@ -1,85 +1,374 @@
...
+
+// Define the evaluation
+Eval('spam-classification', {
+  // Specify which flags this eval uses
+  configFlags: pickFlags('ticketClassification'),
```
c-ehrlich, line 103: This is only defined / explained further down the page. I understand why, and don't really have a better solution, but still feels weird.

ai-engineering/measure.mdx
```diff
@@ -1,85 +1,374 @@
...
+
+    ```ts
+    import { Scorer } from 'axiom/ai/evals';
+
+    const ExactMatchScorer = Scorer(
+      'Exact-Match',
+      ({ output, expected }: { output: string; expected: string }) => {
+        return output === expected ? 1 : 0;
+      }
+    );
```
c-ehrlich: this (or something like it) is what we should be doing in the evals quickstart. showing autoevals there (which is a braintrust product!) makes the code more opaque for no good reason.

ai-engineering/measure.mdx
```diff
@@ -1,85 +1,374 @@
...
 ```bash
-axiom run evals/text-match.eval.ts
+axiom eval --flags-config=experiment.json
 ```
 
-This command will execute the specified test file using `vitest` in the background. Note that `vitest` will be a peer dependency for this functionality.
+## Analyze results in Console
 
-## Analyzing results in the console
+<Badge>Coming soon</Badge> When you run an eval, Axiom AI SDK captures a detailed OpenTelemetry trace for the entire run. This includes parent spans for the evaluation suite and child spans for each individual test case, task execution, and scorer result. Axiom enriches the traces with `eval.*` attributes, allowing you to deeply analyze results in the Axiom Console.
```
c-ehrlich: this doesn't need to be "coming soon", we won't advertise the sdk before the evals view is prod-ready in app

ai-engineering/measure.mdx
```diff
@@ -1,85 +1,374 @@
...
 
-<Badge>Coming soon</Badge> When you run an <Tooltip tip={definitions.Eval}>eval</Tooltip>, the Axiom SDK captures a detailed OpenTelemetry trace for the entire run. This includes parent spans for the evaluation suite and child spans for each individual test case, task execution, and scorer result. These traces are enriched with `eval.*` attributes, allowing you to deeply analyze results in the Axiom Console.
+The results of evals:
+- Pass/fail status for each test case
+- Scores from each scorer
+- Comparison to baseline (if available)
+- Links to view detailed traces in Axiom
+
+Additionally, the AI SDK sends the results to your Axiom dataset for long-term tracking and analysis.
```
c-ehrlich: Don't understand this sentence

c-ehrlich (approved):
Some big picture thoughts:
- I appreciate why we have the "Create/Measure/Observe/Iterate" workflow. But feels strange that there is no page called "Evals" IMO.
- Obviously not part of this PR, but I would LOVE a video at the top of the page

ai-engineering/measure.mdx
```diff
@@ -160,39 +144,48 @@ async function classifyTicket({ subject, content }: { subject?: string; content:
...
-  'Spam-Classification',
-  ({ output, expected }: { 
-    output: { category: string }; 
+  "Spam-Classification",
+  ({ output, expected }: {
+    output: { category: string };
     expected: { category: string };
   }) => {
-    return (expected.category === 'spam') === (output.category === 'spam') ? 1 : 0;
+    const isSpam = (x: { category: string }) => x.category === "spam";
```
c-ehrlich: nit: maybe call this `item` or something instead of `x`

(i know i wrote the original version 😅)

ai-engineering/measure.mdx
```diff
@@ -301,6 +245,14 @@ To run your evaluation suites from your terminal, [install the Axiom CLI](/refer
 | Run eval by name | `axiom eval "spam-classification"` |
 | List available evals without running | `axiom eval --list` |
 
+## Custom scorers
+
+A scorer is a function that scores a capability’s output. Scorers receive the `input`, the generated `output`, and the `expected` value, and return a score.
+
+The `autoevals` library that provides prebuilt scorers for common tasks like semantic similarity, factual correctness, and text matching
```
c-ehrlich: This isn't a full sentence I think? Maybe something like

There are many libraries with prebuilt scorers for common tasks like semantic similarity and text matching that you can wrap, such as `autoevals`.

ai-engineering/measure.mdx
```diff
@@ -301,6 +245,14 @@ To run your evaluation suites from your terminal, [install the Axiom CLI](/refer
...
+
+Use a custom scorer that returns metadata alongside the score:
```
c-ehrlich: Is something meant to come after this?

ai-engineering/measure.mdx
manototh: Yes, sorry, I was still working on this part.

ai-engineering/measure.mdx
manototh: Yes, sorry, I was still working on this part.

c-ehrlich:
Closing in favor of https://github.com/axiomhq/docs/pull/473, feel free to re-open if that's wrong

PR #470, opened 2025-11-17
Update example otel config

Licenser:
Update the example metrics otel config to avoid too large batches


manototh (approved):

manototh:
Thanks @Licenser !

PR #477, opened 2025-11-24
Explain limit behavior with summarize

manototh:
Basis: https://watchlyhq.slack.com/archives/C083K4B8DUY/p1763386581365649


apl/tabular-operators/summarize-operator.mdx
```diff
@@ -178,6 +178,45 @@ Returns a table that shows the heatmap in each interval [0, 30], [30, 20, 10], a
 
 [Run in Playground](https://play.axiom.co/axiom-play-qf1k/query?initForm=%7B%22apl%22%3A%22%5B%27github-push-event%27%5D%20%7C%20where%20_time%20%3E%20ago(7d)%20%7C%20where%20repo%20contains%20%5C%22axiom%5C%22%20%7C%20summarize%20count()%2C%20numCommits%3Dsum(size)%20by%20_time%3Dbin(_time%2C%203h)%2C%20repo%20%7C%20take%20100%22%2C%22queryOptions%22%3A%7B%22quickRange%22%3A%2230d%22%7D%7D)
 
+## Limit behavior with time-binning
+
+When using `limit` or `take` with `summarize` where the first grouping expression is a time bin, Axiom applies the following limit behavior:
+
+1. Compute the global top **N** groups across all time buckets, disregarding the time dimension.
+2. Limit each time bucket to only include those groups that are in the global top **N**.
```
manototh:
```suggestion
1. Limit each time bucket to only include those groups that are in the global top **N**.
```

apl/tabular-operators/summarize-operator.mdx
```diff
@@ -178,6 +178,45 @@ Returns a table that shows the heatmap in each interval [0, 30], [30, 20, 10], a
...
+## Limit behavior with time-binning
+
+When using `limit` or `take` with `summarize` where the first grouping expression is a time bin, Axiom applies the following limit behavior:
+
+1. Compute the global top **N** groups across all time buckets, disregarding the time dimension.
+1. Limit each time bucket to only include those groups that are in the global top **N**.
+
+This means the total number of output rows can be more than **N** rows because each time bucket may contain up to **N** groups. For example, if you have 10 time buckets and limit to 5 groups, you can get up to 50 rows.
+
+### Workarounds
```
manototh: Use collapsible instead of workaround section

apl/tabular-operators/summarize-operator.mdx
```diff
@@ -178,6 +178,47 @@ Returns a table that shows the heatmap in each interval [0, 30], [30, 20, 10], a
...
+    ```
+
+- Convert time to an integer and bin on that instead:
+
+    ```kusto
+    ['sample-http-logs']
+    | extend intTime = toint(_time)
+    | summarize count() by bin(intTime, 3600), status
+    | limit 10
+    ```
```
mhr3, line 218: these workarounds are horrific, wouldn't this do?

```
['http-logs']
| summarize count() by bin(_time, 1m), status
| where isnotnull( _time)
| limit ...
```

manototh: Thanks for this. I took these workarounds from https://app.plain.com/workspace/w_01HA1Z15QQY1X6QR29KFMBMYTH/thread/th_01K9S2045SV355KB2AS6YF39B0/

What you suggest does work: https://play.axiom.co/axiom-play-qf1k/query?qid=2JlSjNZZ5QY-t6g01p&relative=1
@kamstrup Should we only suggest this?

kamstrup: @mhr3 your suggestion works for a very specific use case, but each bucket is still limited implicitly to the global top 1000 groups. It could be our first suggestion, but I think the docs are better if it can suggest a few general mechanisms people can use to work around the problem

manototh: @mhr3 @kamstrup Thanks for your input here. Could you please let me know which workarounds to recommend, if any?

manototh: I would merge it as it is now. If you have any other ideas about this part, please let me know and we'll iterate in another branch.

kamstrup: I still prefer the proposals already on this branch: They might be clunky - but they are workarounds after all - and they work *explicitly* with the quirk described around time-binned-summarize statements. We can iterate in follow-up PRs of course 👍 -- but I still think the focus should be on providing open-ended hints that educate the users about potential solutions, rather a single clear cut solution for a single specific use case.

Michal's proposal with `where isnotnull( _time)` might be simpler, and work *right now*, but ultimately relies on some undocumented quirks in the query execution/compiler toolchain; and does not resolve the restriction to global top-N.

kamstrup (approved):

PR #484, opened 2025-12-01
Add notify on return from no data state

manototh:


Rambatino:
I would potentially add that a threshold breach notification takes precedence over a "Data present" notification.

Rambatino (approved):

monitor-data/threshold-monitors.mdx
```diff
@@ -19,7 +19,7 @@ To create a threshold monitor, follow these steps:
     - The comparison operator is the rule to apply when comparing the results to the threshold. The possible values are **above**, **above or equal**, **below**, and **below or equal**.
     - The frequency is how often the monitor runs. This is a positive integer number of minutes.
     - The range is the time range for your query. This is a positive integer number of minutes. The end time is the time the monitor runs.
-    - **Alert on no data** triggers the monitor when your query doesn’t return any data. Your query returns no data if no events match your filters and an aggregation used in the query is undefined. For example, you take the average of a field not present in any matching events.
+    - **Alert on no data** triggers the monitor when your query doesn’t return any data. Your query returns no data if no events match your filters and an aggregation used in the query is undefined. For example, you take the average of a field not present in any matching events. The monitor sends a notification when data returns after a no-data alert state.
```
manototh:
```suggestion
    - **Alert on no data** triggers the monitor when your query doesn’t return any data. Your query returns no data if no events match your filters and an aggregation used in the query is undefined. For example, you take the average of a field not present in any matching events. The monitor sends a notification when data returns after a no-data alert state. Notifications about crossing the threshold take precedence over notifications about data returning after a no-data state.
```

monitor-data/anomaly-monitors.mdx
```diff
@@ -19,7 +19,7 @@ To create an anomaly monitor, follow these steps:
...
     - The frequency is how often the monitor runs. This is a positive integer number of minutes.
     - The range is the time range for your query. This is a positive integer number of minutes. A longer time range allows the anomaly monitor to consider a larger number of datapoints when calculating the expected value.
-    - **Alert on no data** triggers the monitor when your query doesn’t return any data. Your query returns no data if no events match your filters and an aggregation used in the query is undefined. For example, you take the average of a field not present in any matching events.
+    - **Alert on no data** triggers the monitor when your query doesn’t return any data. Your query returns no data if no events match your filters and an aggregation used in the query is undefined. For example, you take the average of a field not present in any matching events. The monitor sends a notification when data returns after a no-data alert state.
```
manototh:
```suggestion
    - **Alert on no data** triggers the monitor when your query doesn’t return any data. Your query returns no data if no events match your filters and an aggregation used in the query is undefined. For example, you take the average of a field not present in any matching events. The monitor sends a notification when data returns after a no-data alert state. Notifications about deviations from the expected value take precedence over notifications about data returning after a no-data state.
```

monitor-data/threshold-monitors.mdx
manototh: @Rambatino Thanks. Could you please check this again

Rambatino: Great!

PR #487, opened 2025-12-02
Use only AXIOM_DOMAIN placeholder for edges

manototh:
Depends on https://github.com/axiomhq/docs/pull/486

BEFORE MERGING
check for newly added AXIOM_EDGE_DOMAIN placeholders


snippets/replace-domain.mdx
```diff
@@ -1 +1 @@
-Replace `AXIOM_DOMAIN` with `api.axiom.co` if your organization uses the US region. For more information, see [Edge deployments](/reference/edge-deployments).
+Replace `AXIOM_DOMAIN` with `https://us-east-1.aws.edge.axiom.co` if your organization uses the US region, and with `https://eu-central-1.aws.edge.axiom.co` if your organization uses the EU region. For more information, see [Edge deployments](/reference/edge-deployments).
```
dominicchapman: region > edge deployment is a blocker here

might be that we want to avoid including the URLs directly, and have people route through to edge deployment docs to get their URL, but open to keeping it inline until we have more than two. Should refer to edge deployments as, for example, US East 1 (AWS).

dominicchapman (dismissed):

snippets/replace-domain.mdx
manototh: Thanks, good point! Updated

benenharrington (approved):

PR #491, opened 2025-12-03
Fix placeholders in OTel

manototh:


manototh:
No longer necessary after other fixes

PR #517, opened 2026-01-12
Add region param to Vector

manototh:
Preview: https://axiom-mano-region-vector.mintlify.app/send-data/vector#configuration


toppercodes:
Thanks for starting this! I reviewed this against the merged Vector PR [#24037](https://github.com/vectordotdev/vector/pull/24037) and have some feedback:

## Issues

1. **Placeholder value `AXIOM_DOMAIN`** - The `region` field expects a domain-only format like `eu-west-1.edge.axiom.co`, not a generic placeholder. Users won't know what to put here.

2. **Adding `region` to ALL examples** - This parameter is optional. Most users won't need it (they'll use the default cloud endpoint). Adding it everywhere implies it's required.

3. **Missing context** - The PR doesn't explain:
   - What `region` is for (regional edge endpoints for data localization)
   - The expected format (domain-only, e.g., `eu-west-1.edge.axiom.co`)
   - That it's optional
   - Priority order: `url` > `region` > default cloud endpoint

## Suggestions

1. Add a dedicated section explaining the `region` parameter and when to use it
2. Keep most examples without `region` (since it's optional)
3. Add one example specifically demonstrating regional edge usage with a realistic value
4. Document that `url` takes precedence over `region` if both are set

toppercodes:
Also might be worth adding a link or section related to the vector release notes: https://vector.dev/releases/0.52.0/

manototh:
Replied in DM https://watchlyhq.slack.com/archives/D06L91L0LM7/p1768320256112509

toppercodes (approved):

PR #529, opened 2026-01-21
Update align in metrics

manototh:
Basis: https://watchlyhq.slack.com/archives/C0932TPB9UM/p1768450117251899

### DIFF
--- a/query-data/metrics/query-metrics.mdx
+++ b/query-data/metrics/query-metrics.mdx
@@ -30,6 +30,8 @@ To build a typical metrics query:
 1. Click **Run**.
 
 <Note>
+The metric selector only shows metrics that were active during the selected time range. Expand the time range to see more metrics in the dropdown.
+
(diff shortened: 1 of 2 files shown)


query-data/metrics/query-metrics.mdx
```diff
@@ -106,6 +108,8 @@ Available aggregation functions:
 
 For example, to calculate the average over 5-minute time windows, use `align to 5m using avg`. To count the data points in the last hour, use `align to 1h using count`.
 
+When metrics emit data at regular intervals but with different offsets across time series, `align` maps your data to a fixed set of time intervals. This ensures data from different sources is easy to visualize on a single chart. Gaps appear when the alignment interval is shorter than the emission interval. Query over a longer time range or use a longer alignment interval to reduce gaps.
```
benenharrington:
```suggestion
When metrics emit data at regular intervals but with different offsets across time series, `align` maps your data to a fixed set of time intervals. This ensures data from different sources is easy to visualize on a single chart. Null values are returned for intervals where no data points were found, for example if the alignment interval is shorter than the emission interval. By default, non-null values are joined in the resulting chart. You can modify how null values are treated using the chart options menu.
```

benenharrington (approved):

PR #530, opened 2026-01-21
fix: use consistent nested flag paths in CLI examples

bdsqqq:
Update flag override examples in `iterate.mdx` and `analyze-results.mdx` to use nested flag paths (e.g., `--flag.ticketClassification.model`) consistent with the flag schema structure documented in `flags-experiments.mdx`.

The CLI requires flags to match the nested schema structure defined in the app scope. Examples using flat paths like `--flag.model` would fail validation when the flag schema uses nested objects like `ticketClassification.model`.

Fixes AXM-10752

### DIFF
--- a/ai-engineering/iterate.mdx
+++ b/ai-engineering/iterate.mdx
@@ -91,13 +91,13 @@ Use [flags](/ai-engineering/evaluate/flags-experiments) to test different approa
 
 ```bash
 # Test with a more capable model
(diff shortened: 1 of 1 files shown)


bdsqqq:
Closing - the original issue was invalid. All doc examples use valid CLI syntax; different pages show different schema structures which is expected.

PR #543, opened 2026-01-29
Add placeholder config component

manototh:
- Adds config component below code snippets that contain common placeholders.
- Highlights placeholders and replaced values.
- Hides info box below code snippet. If info box contains descriptions of uncommon placeholders, adds the descriptions to the configurator component as text (not as input boxes).
- Keeps info boxes in md file so that LLMs can still read it.

Out of scope: Interactive input box for all placeholders, even uncommon ones like `VECTOR_SOURCE_ID`. This is too risky because there is no safe way to determine if an expression is a placeholder.

Preview:
https://axiom-mano-placeholder-config.mintlify.app/send-data/methods
https://axiom-mano-placeholder-config.mintlify.app/send-data/vector#configuration

Security stance: https://watchlyhq.slack.com/archives/C03HXS7GXGB/p1771506432455879?thread_ts=1771502579.439139&cid=C03HXS7GXGB


guides/opentelemetry-nextjs.mdx
```diff
@@ -48,10 +48,10 @@ OpenTelemetry provides a standardized way to collect and export telemetry data f
         }),
         spanProcessors: [new SimpleSpanProcessor(
           new OTLPTraceExporter({
-            url: `https://${process.env.AXIOM_DOMAIN}/v1/traces`,
+            url: `https://${process.env.AXIOM_URL}/v1/traces`,
```
mhr3: Strictly speaking this is wrong, URL is the whole thing including the scheme, AXIOM_DOMAIN / AXIOM_HOST is correct name for where that env is being substituted into.

manototh: Yes! I was wondering about the right synonym for domain, and host would be perfect. This is especially important because we will want to distinguish between URL and domain, which was discussed during the edge dogfooding session.

But I will only be able to change this where the name of the environment variable is defined in the code snippet. Sometimes the SDK determines this name to be `AXIOM_URL`, I think.

dominicchapman (approved):

PR #551, opened 2026-02-11
Rewrite datetime functions

manototh:


apl/scalar-functions/datetime-functions/dayofyear.mdx
```diff
@@ -0,0 +1,164 @@
+---
+title: dayofyear
+description: 'This page explains how to use the dayofyear function in APL.'
+---
+
+Use the `dayofyear` function in APL to extract the day number within the year from a datetime value. The function returns an integer from 1 to 366 representing how far into the year the given date falls.
```
gordallott: 1 to 356 normally, 1 to 366 in leap years

apl/scalar-functions/datetime-functions/now.mdx
```diff
@@ -0,0 +1,164 @@
+---
+title: now
+description: 'This page explains how to use the now function in APL.'
+---
+
+Use the `now` function in APL to return the current UTC clock time as a `datetime` value, optionally offset by a given timespan. When you use `now` multiple times in the same statement, all instances reference the same clock time.
```
gordallott: should ensure that it's clear that `now()` returns the same time for the entire query, even if it takes 90 seconds to run a query `now()` will always evaluate to the same value

gordallott (approved):
small nits no big deal, mostly just glanced through because it's a lot of diff

PR #583, opened 2026-03-10
Add metrics API reference

manototh:
Preview: https://axiom-mano-mpl-api.mintlify.app/restapi/endpoints/queryMetrics


benenharrington:
It looks like we still show `queryRegion` here. Given that it's a new endpoint, can we only support `queryEdgeDeployment`?

manototh:
> It looks like we still show `queryRegion` here. Given that it's a new endpoint, can we only support `queryEdgeDeployment`?

Good point. It's something that Engineering would need to remove from the API then. This docs is generated based on the axiom repo.

bahlo (dismissed):
this api reference contains fields i'd really not like to have in the public docs, namely:

## query params
- `source` should be done via `accept` header
- `nocache` (does nothing)
- `apl-source` + `apl-source-id` (wrong language)
- totals has the comment to be metricsdb-only which doesn't make sense

## body fields
- `apl` already deprecated, why include this?
- `params` is not something we want to expose yet
- queryOptions is not supported (is it?)
- queryEdgeDeployment

i know this imports existing swagger, so if we can't fix this here this needs a swagger update

benenharrington (approved):

PR #601, opened 2026-03-24
Update popular pages based on do11y

manototh:
Improve navigation and discoverability based on docs analytics

**introduction.mdx:** Replace the Getting started section (which currently has three marketing links) with an intent-based card grid segmented by user goal: send data, query data, monitoring, API, AI engineering, and getting started guides. Users arriving at the highest-traffic page (3,250 entry sessions) currently trigger search at 8.5%. This is the highest rate on the site and the reason can be that the page doesn't direct them to their task.

**restapi/introduction.mdx:** Add a quick-nav card grid at the top of the page linking directly to ingest, query, datasets, and tokens. Add an authentication callout above the API basics section. The page has 2,520 entry sessions and a 23.2% scroll depth, indicating that most users leave before reaching the content they need.

**send-data/methods.mdx:** Move Next.js from "Other methods" to "Popular methods" and reorder rows by actual traffic. Next.js is the 5th highest entry point on the site (594 sessions) but is buried below less-visited integrations.

### DIFF
--- a/introduction.mdx
+++ b/introduction.mdx
@@ -43,6 +43,11 @@ Intuitive web app built for exploration, visualization, and monitoring of your d
 - **Flexible insights:** Real-time query capabilities and an increasingly intelligent UI help pinpoint issues and opportunities without sampling.
 - **AI engineering:** Axiom provides specialized features designed explicitly for AI engineering workflows, allowing teams to confidently build, deploy, and optimize AI capabilities.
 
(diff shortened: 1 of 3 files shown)


manototh:
Comments:
- Can we do slimmer versions of tiles?
- Shouldn't change order of methods
- AI engineering out of place in Intro

PR #612, opened 2026-03-31
Rewrite 29 math functions

manototh:


apl/scalar-functions/mathematical-functions/sqrt.mdx
```diff
@@ -0,0 +1,97 @@
+---
+title: sqrt
+description: 'This page explains how to use the sqrt function in APL.'
+---
+
+Use the `sqrt` function in APL to compute the square root of a non-negative numeric value. The function returns `null` for negative inputs.
```
gordallott: it actually returns `NaN`, which is represented as `null` in the final output. However, we have a `isnan()` function so we should probably indicate that this is a `NaN`

gordallott (approved):
yup looks good :thumbsup: just one small note about a null return mention

apl/scalar-functions/mathematical-functions/sqrt.mdx
```diff
@@ -0,0 +1,126 @@
+---
+title: sqrt
+description: 'This page explains how to use the sqrt function in APL.'
+---
+
+Use the `sqrt` function in APL to compute the square root of a non-negative numeric value.
+
+`sqrt` is useful whenever you want to compress large numeric ranges, compute root-mean-square values, normalize metrics, or undo a squared transformation. It's equivalent to `pow(x, 0.5)` but is more concise for the square-root case.
+
+If the input is negative, `sqrt` returns NaN (Not a Number). APL represents NaN as `null` in query results. Use [isnan](/apl/scalar-functions/mathematical-functions/isnan) to detect these values before aggregating or charting.
```
gordallott:
```suggestion
If the input is negative, `sqrt` returns NaN (Not a Number). APL represents NaN as `null` in query results. Use [isnan](/apl/scalar-functions/mathematical-functions/isnan) to filter these values before aggregating or charting.
```

gordallott (approved):
probably the only thing you would want to do is filter nan's

PR #615, opened 2026-04-03
Update information architecture with metrics and agents

manototh:
Depends on https://github.com/axiomhq/app/pull/7616

## Summary

Restructure the docs information architecture so **logs/traces** and **metrics** are easy to find in Console and in reference, and **AI agents** is a first-class nav group. Align copy with **Builder / Editor** terminology and make **APL vs MPL** responsibilities explicit.

## Navigation changes

- Rename the top-level **Console** group to **Understand data** and nest **Console** (Query, Dashboard, Monitor, Manage, Extend, Intelligence) under it.
- Under **Console → Query**, split content into **Logs and traces** and **Metrics** so readers land on the right signal type quickly. **Canvas** and **Save and export queries** stay at the Query level as shared workflows.
- Add a sibling group **AI agents** under **Understand data** (alongside **Console**) so agent docs are not buried under Intelligence.

image

- On the **Query reference** tab, mirror the same split: an **Overview** entry point, then **Logs and traces (APL)** and **Metrics (MPL)**. MPL reference pages live here next to APL instead of only under main documentation.

*(The tab name stays **Query reference**; it now covers both languages rather than being framed as APL-only.)*

image

### Console / Query

- **`query-data/overview.mdx`** — Onboarding for Datasets, Stream, and Query; Builder vs Editor; how signal types map to query approaches (updated in this branch).
- **`query-data/datasets.mdx`** and **`query-data/stream.mdx`** — Dedicated pages for the other Console tabs alongside Query.
- Refocus **`query-data/explore.mdx`** on querying **logs and traces** in Console (no longer a single catch-all “query everything” page).
- Split former **`explore.mdx`** into:
  - **`query-data/interpret-query-results.mdx`**
  - **`query-data/canvas.mdx`**
  - **`query-data/save-export-query.mdx`**
- **Metrics in Console**
  - **`query-data/metrics/overview.mdx`** — MetricsDB, OTel metrics, when to use MetricsDB vs EventDB.
  - **`query-data/metrics/query-metrics.mdx`** and **`query-data/metrics/query-metrics-mpl.mdx`** — Querying metrics (including MPL in Console).
  - **`query-data/metrics/configure-charts.mdx`** — Chart configuration for metrics.

### Query languages and reference

- **`apl/overview.mdx`** — **Introduction to Axiom query languages**: compares APL vs MPL, links into full APL and MPL reference; sidebar title **Axiom query languages**.
- State clearly across docs that **APL is for logs and traces (EventDB)** and **MPL is for metrics (MetricsDB)**; APL intro cross-links to MPL where relevant.
- **MPL** reference remains **`mpl/introduction`**, **`mpl/sample-queries`**, **`mpl/migrate-metrics`**, now grouped under **Query reference → Metrics (MPL)**.

### Platform copy

- Update **`platform-overview/features`** (and related UI copy) to use **Builder / Editor** instead of legacy Simple/Advanced naming.


dominicchapman (commented):
Observations

- The "Understand data" section is delightful - it elevates AI agents in just the right way - nice one!
- AI engineering now dominates the left sidebar too strongly. I think one helpful resolution here would be to promote use cases out of 'Fundamentals' such that the AI engineering section lives under a 'Use cases' top level section instead, where Observability, AI engineering, and other use cases could live. I'm exploring the following use cases for the site currently: Observability, Log Management, Application Monitoring, AI Engineering, Product Analytics. I think we already mention a few. But might need a wider think. Curious what you think.
- I'm not convinced that separating Logs and traces and Metrics under Query is the best formulation. It creates a separation between these different kinds of machine data (including 'events') which really only matters when it comes to the query language used. I think that can be explained in a more cohesive section. 'Query data' was absolutely the wrong monolith page, but I don't feel good about the separation in the tree as is across machine data types.
- I want to see Logs, traces, and events (APL) - events cannot be skipped as a principal. It shrouds us in narrow engineering 'observability' land. I'm open to something else more concise than L, T, and E.
- Rather than drawing attention to the plurality of languages 'Introduction to Axiom query languages' we could start with a simple intro page that (mostly the same content), but focus on why metrics deserved its own flavor of query language. That's what I care about - the provenance of the language. 
- APL being broken into Functions, Operations, Reference, Migrate feels elegant in the sidebar. But then Metrics 'elements' (name?) is stuffed into MPL > Introduction, which feels inconsistent.

Overall, I like the direction of this a lot. Let me know what you think of the above.

apl/overview.mdx
```diff
@@ -0,0 +1,65 @@
...
+`otel-demo-metrics`:`go.memory.used`
+| where `k8s.deployment.name` == "checkout"
+| align to 5m using avg
+```
+
+[Run in Playground](https://play.axiom.co/axiom-play-qf1k/query?initForm=%7B%22apl%22%3A%22%60otel-demo-metrics%60%3A%60go.memory.used%60%20%7C%20where%20%60k8s.deployment.name%60%20%3D%3D%20%5C%22checkout%5C%22%20%7C%20align%20to%205m%20using%20avg%22%2C%22metricsDataset%22%3A%22otel-demo-metrics%22%7D)
+
+MPL supports filtering by tags, mapping values with functions like `rate` and `fill::prev`, aligning data to time windows, grouping by tag dimensions, and computing across multiple metrics.
+
+If you use PromQL, you can easyily [translate your existing expressions to MPL](/mpl/migrate-metrics).
```
dominicchapman:
```suggestion
If you use PromQL, you can easily [translate your existing expressions to MPL](/mpl/migrate-metrics).
```

manototh:
@dominicchapman These are really good points. Thanks a lot for your insights.

> * AI engineering now dominates the left sidebar too strongly. I think one helpful resolution here would be to promote use cases out of 'Fundamentals' such that the AI engineering section lives under a 'Use cases' top level section instead, where Observability, AI engineering, and other use cases could live. I'm exploring the following use cases for the site currently: Observability, Log Management, Application Monitoring, AI Engineering, Product Analytics. I think we already mention a few. But might need a wider think. Curious what you think.

I love this idea. Implemented in https://github.com/axiomhq/docs/pull/615/commits/ebdb2be90c94fc01f26ec8a6e38bb9d9123aad7b

> * I'm not convinced that separating Logs and traces and Metrics under Query is the best formulation. It creates a separation between these different kinds of machine data (including 'events') which really only matters when it comes to the query language used. I think that can be explained in a more cohesive section. 'Query data' was absolutely the wrong monolith page, but I don't feel good about the separation in the tree as is across machine data types.

I get your point. My rationale was the following:
1. The user experience when querying LTE and metrics is very different.
1. We don't yet have feature parity between LTE and metrics. For example, views, virtual fields, and advanced chart configuration only work for LTE at the moment.

Do you have an alternative in mind? We could separate based on UX first (Builder vs Editor), and signal type second. Would that make more sense to you?

> * I want to see Logs, traces, and events (APL) - events cannot be skipped as a principal. It shrouds us in narrow engineering 'observability' land. I'm open to something else more concise than L, T, and E.

You're right. And it would be nice to find a short version of LTE. But the longer version also works. Implemented in https://github.com/axiomhq/docs/pull/615/commits/31774ba25d6d8fad87f0b9ef947df7f4ee507fb1

> * Rather than drawing attention to the plurality of languages 'Introduction to Axiom query languages' we could start with a simple intro page that (mostly the same content), but focus on why metrics deserved its own flavor of query language. That's what I care about - the provenance of the language.

Is this better? https://github.com/axiomhq/docs/pull/615/commits/1700a8c1e4546b63acb0434071e3f0857625fe0d

> * APL being broken into Functions, Operations, Reference, Migrate feels elegant in the sidebar. But then Metrics 'elements' (name?) is stuffed into MPL > Introduction, which feels inconsistent.

I've changed the metrics intro to `Language features` because it describes the content better. As we expand the MPL docs, we could have a separation similar to APL, but I would do it in another PR.

dominicchapman:
@manototh thanks for this.

> I love this idea. Implemented in ebdb2be

Cool. I suspect we can do more with these use case pages. We probably need to skew ever so slightly more marketing in the language here with time, as an education exercise for folks on why to choose Axiom for that use case over another vendor. 

> Do you have an alternative in mind? We could separate based on UX first (Builder vs Editor), and signal type second. Would that make more sense to you?

In my mind, our desired end state with metrics is one where there is either feature parity across machine data types, or an explicit understanding that some feature is not and will not be supported. As a reader, I expect I’d prefer to go to one spot for understanding views, for example, is a feature of Console, and then see in a simple summary table that the feature is either not supported, or not yet supported, with plans to support. Perhaps I'm not understanding the size of the feature gap across different machine data types, though.

> Is this better? 1700a8c

I think this is strong! 👍 

> As we expand the MPL docs, we could have a separation similar to APL, but I would do it in another PR. 

Great stuff.

manototh:
Close in favour of https://github.com/axiomhq/docs/pull/632

PR #621, opened 2026-04-13
Update CloudWatch forwarder docs to v1.3.0 with EU edge support

thecraftman:
The docs referenced outdated v1.1.1 CloudFormation templates which lack EU region support. Updates all stack URLs to v1.3.0 and adds documentation for the AxiomEdge and AxiomEdgeURL parameters needed for EU deployments.

### DIFF
--- a/send-data/cloudwatch.mdx
+++ b/send-data/cloudwatch.mdx
@@ -40,9 +40,29 @@ To install the Axiom CloudWatch Forwarder, choose one of the following:
 
 ### Install with Cloudformation stacks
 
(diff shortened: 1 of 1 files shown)


kasperaamodt:
@thecraftman BTW, when I was trying to get this to work yesterday, I saw v1.1.1 in the template url, and manually changed it to v1.3.0. When I did that, the stack deployment failed in a later stage, because the Lambda zip was hosted in `us-east-1`, while I was trying to deploy the stack in `eu-central-1`.

PR #634, opened 2026-04-23
Specify what term means

### LINKED ISSUE #602: Issue on docs
Path: /apl/scalar-operators/string-operators

Does not explain how string is split into terms. E.g. `foo-bar` I presume is two terms, `foo` and `bar`. What about `foo_bar`?

manototh:
Basis: axiom/pkg/kirby/query/filter/match_str.go


dominicchapman (approved):

kamstrup (approved):
Yep, this is accurate 👍

gordallott (approved):

PR #657, opened 2026-06-10
feat: docs for correlations

c-ehrlich:
adds basic correlation docs

could spice up with screenshots etc

but best to wait until after the feature lands


query-data/correlations.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+Correlations connect datasets that describe the same system from different telemetry signals. A correlation group records which logs, traces, and metrics datasets belong together.
+
+Use correlations to move between related signals without manually copying IDs or rebuilding context in another query.
+
+## What correlations are
+
+A correlation group contains up to three datasets:
+
+- **Logs dataset:** Events that contain log records.
```
Licenser: should we at least hint to this ideally being OTEL logs? It might be missleading when people think they can correlate abrbitary logs w/o any rules on them for correlation

query-data/correlations.mdx
```diff
@@ -0,0 +1,77 @@
...
+Axiom uses the datasets in a correlation group to find related data during an investigation. For example, a log event that contains a trace ID links to the matching trace. A trace links back to log records from the correlated logs dataset. A span in a trace links to metrics from the correlated metrics dataset.
+
+## Prerequisites
+
+Before you create a correlation group, prepare the datasets that you plan to connect:
+
+- Send logs, traces, or metrics to separate datasets.
+- Use a dedicated OpenTelemetry metrics dataset for metrics.
+- Include trace identifiers in logs for log-to-trace and trace-to-log navigation.
+- Put the metrics dataset on an edge deployment to use trace-to-metrics correlation.
```
Licenser: should we guide people to put all datasets on the same edge? Other wise they'll probably run into problems right?

query-data/correlations.mdx
```diff
@@ -0,0 +1,77 @@
...
+1. In Axiom Console, go to **Datasets**.
+1. Click **Correlations**.
+1. Click **New correlation group**.
+1. Enter a **Name**.
+1. Enter a **Slug**.
+1. Select one or more datasets:
+   - **Logs dataset**
+   - **Traces dataset**
+   - **Metrics dataset**
+1. Click **Create correlation group**.
```
Licenser, line 41:
```suggestion
1. In Axiom Console, go to **Datasets**.
2. Click **Correlations**.
3. Click **New correlation group**.
4. Enter a **Name**.
5. Enter a **Slug**.
6. Select one or more datasets:
   - **Logs dataset**
   - **Traces dataset**
   - **Metrics dataset**
7. Click **Create correlation group**.
```

query-data/correlations.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+Axiom uses the datasets in a correlation group to find related data during an investigation. For example, a log event that contains a trace ID links to the matching trace. A trace links back to log records from the correlated logs dataset. A span in a trace links to metrics from the correlated metrics dataset.
+
+## Prerequisites
+
+Before you create a correlation group, prepare the datasets that you plan to connect:
+
+- Send logs, traces, or metrics to separate datasets.
+- Use a dedicated OpenTelemetry metrics dataset for metrics.
+- Include trace identifiers in logs for log-to-trace and trace-to-log navigation.
```
Licenser: how does that look like? just any field that includes a trace id? How about span id's? Can we give an example what works best

Licenser (approved):
Yaaaassss! Just some nit picks

query-data/correlations.mdx
c-ehrlich: numbering is intentionally `1.` for every line, the build step figures it out

query-data/correlations.mdx
c-ehrlich: updated here: https://github.com/axiomhq/docs/blob/correlation-docs/query-data/correlations.mdx?plain=1#L27

query-data/correlations.mdx
c-ehrlich: I don't think different edges actually cause any problems with the current setup. We never query multiple datasets in the same query. What issues do you see?

query-data/correlations.mdx
c-ehrlich: good point, cleaned up: https://github.com/axiomhq/docs/blob/correlation-docs/query-data/correlations.mdx?plain=1#L13-L17

Licenser (approved):
re-approve after change :D woooh!

PR #669, opened 2026-07-06
docs: document dataset schema locking

benenharrington:
## What

Documents the new dataset schema locking feature as a console workflow.

- **reference/datasets.mdx**: new [Lock dataset schema](https://axiom.co/docs/reference/datasets#lock-dataset-schema) section alongside the trim and vacuum sections. Covers what locking does (snapshots the schema — fields **and their types** — over the retention window; off-schema fields and type-mismatched field values are discarded at ingest while events are still ingested), that it's instant and reversible without modifying stored data, and the constraints (event datasets only, requires datasets update permission, not available on shared datasets). Includes lock and unlock steps via the Fields panel overflow menu.
- **reference/limits.mdx**: links schema locking from the datasets/fields restrictions section as a way to prevent new fields from being added.
- **query-data/datasets.mdx**: mentions the default add-fields-at-ingest behavior and the locked exception in the Datasets tab overview, linking to the reference section.

## Notes

- No API reference docs: the lock/unlock endpoints are internal-only for now, so this is documented as a console workflow only.
- Vale passes on all three files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/reference/datasets.mdx
+++ b/reference/datasets.mdx
@@ -140,6 +140,38 @@ To vacuum fields, follow these steps:
 1. In the list, find the dataset where you want to vacuum fields, and then click <img src="/doc-assets/icons/vacuum.svg" className="inline-icon" alt="Vacuum fields icon" /> **Vacuum fields** on the right.
 1. Select the checkbox, and then click **Vacuum**.
 
+## Lock dataset schema
+
+By default, the data schema of a dataset grows together with your data. When you ingest an event with a field that isn’t yet part of the data schema, Axiom adds the new field to the schema.
+
+Locking the schema of a dataset freezes its set of fields. When you lock the schema, Axiom takes a snapshot of the fields that occur in events within your retention period. While the schema is locked, Axiom discards fields that aren’t part of this snapshot during the data ingestion process. Axiom still ingests the events themselves and only drops the fields that aren’t part of the locked schema.
+
+Locking the schema can be useful if you want to protect a dataset with a well-defined schema from unexpected fields, or if you want to keep the number of fields within the [allowed limits](/reference/limits#pricing-based-limits). To reduce the number of fields already associated with a dataset, [vacuum its fields](#vacuum-fields).
+
+Locking and unlocking the schema take effect immediately and don’t change the data stored in the dataset. When you unlock the schema, Axiom restores the default behavior and adds new fields to the schema during the data ingestion process. Unlocking doesn’t restore fields that Axiom discarded while the schema was locked.
+
+<Note>
+- You can only lock the schema of event datasets. Metrics datasets don’t support schema locking.
+- To lock or unlock the schema, you need update permissions for datasets. Without these permissions, Axiom doesn’t display the menu options below.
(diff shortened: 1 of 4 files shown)


reference/datasets.mdx
```diff
@@ -140,6 +140,63 @@ To vacuum fields, follow these steps:
 1. In the list, find the dataset where you want to vacuum fields, and then click <img src="/doc-assets/icons/vacuum.svg" className="inline-icon" alt="Vacuum fields icon" /> **Vacuum fields** on the right.
 1. Select the checkbox, and then click **Vacuum**.
 
+## Lock dataset schema
+
+By default, the data schema of a dataset grows together with your data. When you ingest an event with a field that isn’t yet part of the data schema, Axiom adds the new field to the schema.
```
SayedAlesawy: suggestion: Similarly if you ingest a different type of value for a field that already exists, Axiom extends that field to support both existing and new types.

reference/datasets.mdx
```diff
@@ -140,6 +140,63 @@ To vacuum fields, follow these steps:
...
+
+Locking the schema of a dataset freezes its set of fields and the type of each field. While the schema is locked, Axiom discards the following during the data ingestion process:
+
+- Fields that aren’t part of the locked schema.
+- Field values whose type doesn’t match the type of the field in the locked schema. For example, if the locked schema defines a field as a number and you ingest an event where the field contains a string, Axiom drops the field’s value from that event.
```
SayedAlesawy: suggestion: ... whose type doesn’t match the type of the field in the locked schema and can't be promoted to match it. 

And in the example we can mention that for instance an int can go into a float field normally. Also a null can go into field.

reference/datasets.mdx
```diff
@@ -140,6 +140,63 @@ To vacuum fields, follow these steps:
 1. In the list, find the dataset where you want to vacuum fields, and then click <img src="/doc-assets/icons/vacuum.svg" className="inline-icon" alt="Vacuum fields icon" /> **Vacuum fields** on the right.
 1. Select the checkbox, and then click **Vacuum**.
 
+## Lock dataset schema
```
SayedAlesawy, line 145: I think we also need to change docs for map field creation and say that creating a map field will regenerate the locked schema if a dataset is locked and then link to this section?

reference/datasets.mdx
benenharrington: Added in cedb082e — the default-behavior paragraph now covers type extension as well as new fields.

reference/datasets.mdx
benenharrington: Added in cedb082e — the bullet now says values are only dropped when they can't be promoted, with the integer-to-float and null examples.

reference/datasets.mdx
benenharrington: Added in cedb082e — the map fields page now notes that map field changes regenerate a locked schema, linking back to this section. Also documented the same re-snapshot behavior for vacuum.

SayedAlesawy (approved):

PR #675, opened 2026-07-15
docs: agent-created organizations

benenharrington:


content/docs/(documentation)/console/intelligence/agent-created-orgs.mdx
```diff
@@ -0,0 +1,105 @@
+---
+title: Agent-created organizations
+description: 'This page explains how AI agents can provision a temporary Axiom organization with a single API request, and how a human can later claim the organization and keep everything the agent built.'
+sidebarTitle: Agent-created orgs
```
dominicchapman, line 4: I wonder if this reads like an unusual resource category alongside the other entries. Is "Provision org" better?

content/docs/(documentation)/console/intelligence/agent-created-orgs.mdx
```diff
@@ -0,0 +1,105 @@
...
+keywords: ["ai agents", "agent", "provision", "temporary organization", "claim", "org", "bootstrap", "mcp"]
+---
+
+An AI agent can create a real, fully functional Axiom organization with a single unauthenticated HTTP request. The response includes an API token the agent can use immediately to create datasets, ingest and query data, and build dashboards and monitors.
```
dominicchapman, line 8: Is fully functional accurate when there are reduced quotas and notifications are disabled? Perhaps fine; I get the idea.

console/intelligence/agent-created-orgs.mdx
```diff
@@ -0,0 +1,105 @@
...
+}
+```
+
+| Field | Description |
+| :--- | :--- |
+| `id` | The organization ID. Use it in API routes that require an org ID. |
+| `name` | The display name of the organization. |
+| `defaultEdgeDeployment` | The edge deployment where the organization stores its data. |
+| `expiresAt` | The time at which the organization is deleted if it hasn’t been claimed. This is 24 hours after provisioning. |
+| `claimUrl` | The URL a human opens to take ownership of the organization. Surface this URL to the user. |
```
dominicchapman: I suppose this is similarly sensitive?

content/docs/(documentation)/console/intelligence/agent-created-orgs.mdx
```diff
@@ -0,0 +1,105 @@
...
+2. Sign in to Axiom. If you don’t have an Axiom account yet, one is created as part of the flow.
+3. You become the owner of the organization and the organization becomes permanent.
+
+Claiming keeps everything the agent built: datasets, ingested data, dashboards, monitors, and other configuration all carry over unchanged.
+
+Claiming also has the following effects:
+
+- The temporary markers are removed. The organization no longer expires and behaves like any other Axiom organization on the Personal plan.
+- The throttled limits are lifted and the full [Personal plan limits](/reference/limits) apply.
+- The synthetic agent user is removed. The API token issued at provisioning continues to work, so the agent keeps its API access. As the new owner, you can revoke it or create additional [API tokens](/reference/tokens) at any time.
```
dominicchapman, line 83: Who does this token belong to from an audit perspective?

content/docs/(api-reference)/restapi/versions/v2.json
```diff
@@ -1333,6 +1333,39 @@
...
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "#/components/schemas/ProvisionOrg"
+              }
+            }
+          },
+          "required": true
+        },
+        "responses": {
```
dominicchapman, line 1355: Are we missing 429?

dominicchapman (approved):
Solid material. Great work!
Added comments for things that stood out to me, but nothing I consider blocking.

content/docs/(documentation)/console/intelligence/agent-created-orgs.mdx
benenharrington: I think "Provision org" might stand out a bit as a verb / action. "Org creation" would fit but it might be confusing if a regular user lands on that page and thinks it's the typical org creation flow.

content/docs/(documentation)/console/intelligence/agent-created-orgs.mdx
benenharrington: I think it's OK. Personal tier orgs are 'fully functional' but don't have the same set of notifiers available as paying orgs

console/intelligence/agent-created-orgs.mdx
benenharrington: I'm missing a related message I think. Is the idea that the description should make it clear that the field value is sensitive? And similarly for the token returned? That makes sense

content/docs/(documentation)/console/intelligence/agent-created-orgs.mdx
benenharrington: Good q that I needed to go check the answer to. Nothing changes about how it appears in the audit log before / after claim. actor.email is null and actor.creator_id is a UUID
Screenshot 2026-07-28 at 13 48 31

content/docs/(api-reference)/restapi/versions/v2.json
benenharrington: I'm not sure why it's not in the swagger that generates this file, it does exist. I'll add it 👍

console/intelligence/agent-created-orgs.mdx
dominicchapman: Yep. Further down in page it's alluded to in relation to `token` but suspect `claimUrl` should also be handled delicately

content/docs/(api-reference)/restapi/versions/v2.json
benenharrington: (claude) Good catch — we do return 429, from two places: the per-IP rate limit middleware and the global daily provisioning cap. It was missing here because the source swagger only declared the 200. Fixed upstream in axiomhq/axiom#18300 and mirrored in this branch (3d651000) so the next spec sync doesn't drift.

PR #726, opened 2026-09-09
Add MPL guidance for querying summary metrics

syoungl:
summary metrics are rare, legacy but there are users using it, but it doesn't have dedicated bucket function like interpolate_***_histogram (and we don't plan to), so it needs more guidance 

- Add a Query reference page covering both: use a bucket function for histograms, and select components with the reserved axiom tags plus compute for summaries. 

- Also add a bucket distribution sample query (which we use often in metrics dashboard!)


benenharrington (approved):
Nice!

syoungl:
thanks for review !

